PR #4, opened 2016-07-04
Water Quality Data Model

joaquincabezas:
Contribution from Adevice


WaterQuality/WaterQualityObserved/doc/spec.md
```diff
@@ -0,0 +1,182 @@
+# Water Quality 
+
```
jmcanterafonseca: this should be entitle Water Quality Observed

jmcanterafonseca:
I don't know why the full spec of WaterQualityObserved appears as new, please check

PR #7, opened 2016-08-15
updated GeoJSON Normative References for WasteManagement

cayetanobv:
Dear team,

I suggest the following change on the basis of the recently released RFC 7946 for GeoJSON format by the IETF (2016-08-10): https://datatracker.ietf.org/wg/geojson/documents/

In the coming months GeoJSON format probably already become standard, so there will be new changes. I think it's better to put the link to IETF datatracker (instead a direct link to last documents) where is the status of the format and the link to the last document.

I did not change any more that line 15 from both files but Git have been tracked as changes (and there are no more changes at all) the change of tabs for spaces (or something like that).

Cayetano Benavent,
GEOGRAPHICA

### DIFF
--- a/WasteManagement/WasteContainer/doc/spec.md
+++ b/WasteManagement/WasteContainer/doc/spec.md
@@ -6,37 +6,37 @@ A waste container.
 
 ## Data Model
 
-+ `id` : Unique identifier. 
++ `id` : Unique identifier.
 
 + `type` : Entity type. It must be equal to `WasteContainer`.
 
-+ `location` : Container's location represented by a GeoJSON Point. 
++ `location` : Container's location represented by a GeoJSON Point.
     + Attribute type: `geo:json`.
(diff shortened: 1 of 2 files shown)


jmcanterafonseca:
thanks for the contribution. I would prefer a direct link to the spec rather to an intermediate resource, it is clearer.  A link to https://tools.ietf.org/html/rfc7946 would suffice? what do you think?

On the other hand please could you configure properly your editor, so that only the needed changes are commited?

thanks!

cayetanobv:
@jmcanterafonseca I think your solution is perhaps clearer. I will upload the changes again (and remove the remaining "no" changes).

cayetanobv:
@jmcanterafonseca I updated code with change as you proposed.

The issue with the  "no changes" tracked by Git is an Atom trailing spaces issue (a default True value for property "Remove trailing spaces"): https://github.com/atom/whitespace/issues/10

Thanks!

jmcanterafonseca:
LGTM, thanks

PR #19, opened 2016-09-20
FIX OffStreetParking. Remove the concept of implicit group and make g…

jmcanterafonseca:
…roups explicit when there is no parking homogeneity


jmcanterafonseca:
@gtorodelvalle r?

Issue #10, opened 2016-09-09
rename `dateUpdated` to `timestamp` to comply with GSMA's `ExtQuantitativeValue`


jmcanterafonseca:

jmcanterafonseca:
fixed

PR #31, opened 2016-09-23
`dateUpdated`  --> `timestamp`

jmcanterafonseca:
please note that at entity level  `dateUpdated` is now `dateModified` as per NGSIv2 that's why a small number of renames are to `dateModified`


jmcanterafonseca:
@gtorodelvalle r?

jmcanterafonseca:
solves #10

jmcanterafonseca:
yes, because the right property is dateModified as per NGSIv2

PR #72, opened 2016-10-10
Includes powerState attr when no group is defined

javimudi:
Hi, team,

We're proposing this attribute to handle power state for groupless cabinets.

When no group of streetlights is been logically managed, including an empty group just to handle the power state of the circuits under the control cabinet seems like an unneeded overhead.

Thus, we think that allowing an empty group as well as including the powerState attribute could be an improved approach.

Regards


jmcanterafonseca:
This change worries me. If we move powerState to the cabinet, we would need to move other attributes at StreetlightGroup level related to it, which would make the model messy.

jmcanterafonseca:
closing as there is no activity and no agreement

Issue #76, opened 2016-10-25
Would be nice to have a json schema for each data model


LeonanCarvalho:
The documentation is easily readable to humans. Thinking about machine-readable concept would be nice to have a file containing the JSON data model schema because markdown isn't a good structure to do that, anyway using this format could be easily consumed by developers also. 

Despite having a pretty good [documentation](https://www.fiware.org/data-models/), references and links, there is no kind of support to developers or ways to automatize and consumers of this standardize the data models as schemes or another way to be a machine-redabe model.
Data Models there's no limits, new models and changes on currently ones will be necessary along time. It  will evolve aligned to new discoveries, news technologies. So, I think Data Models should be flexible, with version control and scheme standard files on a own repository to be consumed or "learned" by smart applications and developers through a know semantic.

Extracting from the docs, I could build a scheme with all attributes of the Device Data Model. I tried to cover all specs to be consumed by an application. The result was:

``` json
{
    "namespace": "Environment",
    "name": "WaterQualityObserved",
    "attributes": {
        "id": {
            "descr": "Unique identifier.",
            "type": "Text",
            "mandatory": true
        },
        "type": {
            "descr": "Entity type. It must be equal to WaterQualityObserved.",
            "type": "Text",
            "allowedvalues": "WaterQualityObserved",
            "mandatory": true
        },
        "location": {
            "descr": "Location where measurements have been taken, represented by a GeoJSON Point.",
            "type": "geo:json",
            "allowedvalues": {
                "normative": "https://tools.ietf.org/html/rfc7946"
            },
            "mandatory-if": [
                {
                    "attribute": "address",
                    "rule": "equals",
                    "value": "undefined"
                }
            ]
        },
        "address": {
            "descr": "Civic address where the Water Quality measurement is taken.",
            "type": "address",
            "allowedvalues": {
                "normative": "https://schema.org/address"
            },
            "mandatory-if": [
                {
                    "attribute": "location",
                    "rule": "equals",
                    "value": "undefined"
                }
            ]
        },
        "temperature": {
            "descr": "Temperature.",
            "type": "Number",
            "metadata": {
                "timestamp": {
                    "type": "DateTime"
                }
            },
            "unit": {
                "symbol": "CÂº",
                "descr": "Celsius Degrees"
            }
        },
        "conductivity": {
            "descr": "Electrical Conductivity.",
            "type": "Number",
            "metadata": {
                "timestamp": {
                    "type": "DateTime"
                }
            },
            "unit": {
[...]

PR #85, opened 2016-10-31
Include VehicleModel Schema

LeonanCarvalho:
regarding to issue #76


jmcanterafonseca:
closing. We will be developing schemas for the data models but they will be based on standard JSON-Schema

PR #89, opened 2016-12-01
ADD harvest process for weather forecast

jmcanterafonseca:


jmcanterafonseca:
just landing. new code for harvesting weather forecast

PR #99, opened 2017-01-31
Water quality schema. Fix data model adding `measurand`as attribute and clarifying usage

jmcanterafonseca:


jmcanterafonseca:
@gtorodelvalle r?

PR #100, opened 2017-02-01
Waste Container Schema. Refining id character restriction

jmcanterafonseca:


jmcanterafonseca:
@gtorodelvalle r?

@fgalan can you have a look at the reg exp for ids?

thanks

common-schema.json
```diff
@@ -3,12 +3,17 @@
   "id": "https://fiware.github.io/dataModels/common-schema.json",
   "title": "Common definitions for FIWARE Harmonized Data Models",
   "definitions": {
+    "EntityIdentiferType": {
+      "type": "string",
+      "minLenght": 1,
```
kzangeli: Typo: minLenght    ... th

common-schema.json
```diff
@@ -3,12 +3,17 @@
...
+      "maxLength": 256,
+      "pattern": "^[\\w\\-:\\.,;\\(\\)/\\{\\}]+$"
```
fgalan: Two comments:

* Check that "all symbols in the ACCII basic set" minus `[&, ?, /. #]` is equal to the symols added to `\w` in this line
* This regex is aligned with the rules at NGSIv2 spec for ID? Or with the rules at NGSIv2 spec for ID + specific Orion implementation rules (that adds additional forbidden chars to the NGSIv2 spec basic set)?

fgalan: @jmcanterafonseca offline confirms that is the second case, so the regex should also forbid the following chars:


*    <
*    >
*    "
*    '
*    =
*    ;
    (
    )

common-schema.json
jmcanterafonseca: well spotted, unfortunately JSON Schema validators didn't warn about that

fgalan: Isn't there any JSON Schema to validate the JSON Schema itself :)?

jmcanterafonseca: fixed https://github.com/Fiware/dataModels/pull/100/commits/a6b3f5b43a2d3ffb11779abd01a8f58eca951324

common-schema.json
jmcanterafonseca: fixed https://github.com/Fiware/dataModels/pull/100/commits/a6b3f5b43a2d3ffb11779abd01a8f58eca951324

gtorodelvalle:
Merging after 2 LGTMs... :)

PR #112, opened 2017-04-14
Added Travis CI for testWeatherObserved

chicco785:
This set of commits provide Continuous Integration test for Json Schemas and their examples. At the moment the only CI included is testWeatherObserved. Further will come.


jmcanterafonseca:
several comments:

A/ There are some files which appear as modified but they shouldn't. Please could you check the Pull Request. It seems you have an issue with your editor. 

B/ It seems that the testing script could be made generic and do the testing for each of the data models by iterating over the different folders which actually contain data models. Using that approach it would be very easy to add and test new data models. What do you think?

chicco785:
I will check the issue with the editor. Regarding the automation, it could be possible if all JSON Schema that we import from external sources are correctly de-referenciable. Unfortunately, this appears to not be the case, not only for the geo schema used by FIWARE, but also for other "standard" JSON Schemas :(

README.md
```diff
@@ -39,14 +39,51 @@ A `validate.sh` script is provided for convenience.
...
+
 ## How to contribute
 
-Contributions should come in the form of pull requests. 
+Contributions should come in the form of pull requests.
+
+New data models should include:
+- A data model description based on the [data model template](https://github.com/Fiware/dataModels/blob/master/datamodel_template.md), e.g. [spec.md](https://github.com/Fiware/dataModels/blob/master/Weather/WeatherObserved/doc/spec.md) (model specifications should be stored in the doc folder of the data model)
+- A summary file (as an extract from the spec file), e.g. [README.md](https://github.com/Fiware/dataModels/blob/master/Weather/WeatherObserved/README.md)
+- A JSON Schema definition, e.g. [schama.json](https://github.com/Fiware/dataModels/blob/master/Weather/WeatherObserved/schema.json)
```
chicco785: fixed

README.md
```diff
@@ -39,14 +39,51 @@ A `validate.sh` script is provided for convenience.
...
+- A JSON example file, e.g.  [example.json](https://github.com/Fiware/dataModels/blob/master/Weather/WeatherObserved/example.json)
+
+### Continuous integration
+To facilitate the validation of changes on existing models, and ensuring the validity of new ones, we introduced continuous integration tests. Pull requests will require the passing of continuous integration tests.
+
+Contributors of new data models, should
+- Create a new set of tests for their models and store in it in the ```test``` folder as shell script;
```
chicco785: fixed

chicco785:
A/fixed

B/as commented above. @taliaga actually developed such a script, but in case of referenced JSON Schema that need to be manually imported, the validation will fail.

C/fixed (at least i hope so)

README.md
```diff
@@ -39,14 +39,63 @@ A `validate.sh` script is provided for convenience.
...
+    - `README.md`
+    - `schema.json`
+    - `example.json`
+  - `NewModelEntityTwo/`
+    - `doc/`
+      - `spec.md`
+    - `README.md`
+    - `schema.json`
+    - `example.json`
+
```
jmcanterafonseca: too many empty lines

jmcanterafonseca:
I still think it is better to bind tests to data models by creating a test folder inside each data model.

README.md
```diff
@@ -39,14 +39,63 @@ A `validate.sh` script is provided for convenience.
 
 Note: JSON Schemas only capture the NGSI simplified representation (`options=keyValues`)
 
+
```
jmcanterafonseca: Are these new empty lines really needed?

jmcanterafonseca:
there are still changes that shouldn't be, please check your editor

README.md
chicco785: it looks like that in some part of the readme there where white trailing spaces, that Atom automatically remove, I will try to disable that. Even though, in general, I agree with Atom :) there should not be trailing spaces at the end of a line.

PR #114, opened 2017-04-17
refactor: Schemas removed

netzahdzc:
Those schemas are removed since they have been either integrated into
respective schemas or are been referenced to respective ohm schema.


netzahdzc:
sorry, I accidentally send this pull request

PR #123, opened 2017-05-10
New fields

VictorVicente1:
Introduction of the new fields priceRatePerMinute y priceCurrency to display the price rate per minute and the price currency


jmcanterafonseca:
this PR should be collapsed with #124. Please use git cherry-pick to do so.

PR #124, opened 2017-05-10
New fields OffStreetParking : pricePerMinute

VictorVicente1:
Introduction of the new fields priceRatePerMinute y priceCurrency to display the price rate per minute and the price currency

### DIFF
--- a/Parking/OffStreetParking/doc/spec.md
+++ b/Parking/OffStreetParking/doc/spec.md
@@ -97,6 +97,15 @@ a resident and a disabled permit are needed to park. If empty or `null`, no perm
     + Normative references: https://schema.org/acceptedPaymentMethod
     + Optional
     
++ `priceRatePerMinute` : Price rate per minute.
+    + Attribute type: [Number](https://schema.org/Number)
+    + Optional
+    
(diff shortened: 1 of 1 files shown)


jmcanterafonseca:
thanks for this contribution. it would be better if you collapse the two pull requests into one as we need to land at the same time the json schema and the spec.

Parking/OffStreetParking/doc/spec.md
```diff
@@ -97,6 +97,15 @@ a resident and a disabled permit are needed to park. If empty or `null`, no perm
     + Normative references: https://schema.org/acceptedPaymentMethod
     + Optional
     
++ `priceRatePerMinute` : Price rate per minute.
+    + Attribute type: [Number](https://schema.org/Number)
+    + Optional
+    
++ `priceCurrency` : Price currency of price rate per minute.
+    + Attribute type: [Text](http://schema.org/Text)
+    + Normative references: https://schema.org/priceCurrency  https://schema.org/priceCurrency
```
jmcanterafonseca: this has to be changed by [https://schema.org/priceCurrency](https://schema.org/priceCurrency)

PR #136, opened 2017-06-07
Remove eol characters from descriptions

aarranz:
Seems that current versions of the Orion Context Broker is not able to handle EOL characters inside attribute values.


jmcanterafonseca:
LGTM, thanks

PR #140, opened 2017-09-14
Refactor air quality observed to remove measurand and support qualita…

jmcanterafonseca:
…tive values


jmcanterafonseca:
@sergg75 r?

PR #143, opened 2017-09-29
Added entity VehicleItinerary to schedule routes for Vehicles

AnderPijoan:
Introduction of the VehicleItinerary entity in the transportation model to be able to define itineraries for vehicles. It contains a list of ordered stops at other entities the vehicle should follow together with the estimated arrival and departure time.

This will be used in Waste4Think [1] project for scheduling everyday collection routes of recycling trucks.

[1] http://waste4think.eu/


jmcanterafonseca:
general comment:

This is strange that an itinerary is defined in terms of city assets, containers, etc.  It would limit reusability, imagine that we need to do the same for every asset in a city ... 

An itinerary should be defined in terms of the different geographical points to be visited ...

AnderPijoan:
That is a discussion we also had. Itineraries always have a purpose. An itinerary is not just traveling to a certain point, there is some intention of interacting there with some entity:

+ Gathering passengers at bus stops.
+ Delivering packages at houses, business, depots.
+ Collecting waste containers. 
+

So thats why we think it is more meaningful and explicit to keep a reference to the entities you need to visit. The entities' location will be contained in that entity's instance. I can not think of any use case in which a vehicle should just travel to some locations without interacting there.

jmcanterafonseca:
please DO NOT modify in the same PR different data models. As per our way of working one PR should correspond to one change, so if this PR was devoted to introduce a new data model, stick to it and not change other data models

jmcanterafonseca:
I think a better design can be devised. 

We can have the following entity types:

`Route`, agnostic about the vehicle which is going to be assigned to such route. A route will be represented by a MultiLineString in GeoJSON. A route can have a name, description, or even the service to which the route is related to. 

`Mission` which will represent what's the mandate of a vehicle. A Mission will be composed by multiple `MissionItem` entities each one indicating the concerned asset, the service to be executed over the asset and the scheduled dates. 

`Vehicle` we can add an extra property to vehicle which is the current mission assigned to it.

Parking/ParkingAccess/doc/spec.md
```diff
@@ -14,7 +14,7 @@ Represents an access point to a parking site, normally an offstreet parking.
     + Attribute type: `geo:json`.
     + Normative References: [https://tools.ietf.org/html/rfc7946](https://tools.ietf.org/html/rfc7946)
     + Mandatory
-    
+
```
jmcanterafonseca: please do not modify data models without substantive changes. please stick to the PR

Transportation/Vehicle/VehicleItinerary/doc/spec.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Example
+
+    {
+      "id": "vehicleitinerary:89237",
+      "type": "VehicleItinerary",
+      "name": "VehicleItinerary 12",
+      "description": "Bus Vehicle Itinerary for 2017-09-22",
+      "stops": [
```
jmcanterafonseca: this is not proper NGSI design. please stick to simple (attribute, value) data structures. please check my comment about an alternative design.

Transportation/Bike/BikeHireDockingStation/doc/spec.md
```diff
@@ -124,24 +119,24 @@ A JSON Schema corresponding to this data model can be found [here](http://fiware
...
-	   "freeSlotNumber": 10,
-	   "address": {
-		  "streetAddress": "Paseo Antonio Banderas (Diputación)",
-		  "addressLocality": "Malaga",
-		  "addressCountry": "España"
-	   },
-	   "description": "Punto de alquiler de bicicletas próximo a Diputación",
-	   "dateModified": "2017-05-09T09:25:55.00Z"
-	}
+    {
```
jmcanterafonseca: please do not make not substantive changes to existing data models.

jmcanterafonseca:
as we agreed during last TSC it would be great that the SmartSDK team (@chicco785) could start giving feedback to PRs made over this codebase

Transportation/Bike/BikeHireDockingStation/doc/spec.md
AnderPijoan: Sorry, this change was in for the docs to pass the checks.

Parking/ParkingAccess/doc/spec.md
AnderPijoan: Sorry, this change was in for the docs to pass the checks.

chicco785:
hi, ubiwhere worked on a similar case (base on google transit feeds) for public transports (https://github.com/ftcardoso/dataModels/tree/PublicTransportation/PublicTransportation).  i think it would be beneficial to converge. feelings?

few comments on the current discussion:
* i think stops could be mapped as POI or something similar
* I agree with Jose, the same itinerary (in term of stops) may have different schedules, it may be worth to avoid duplicating such information, by creating a Route concept (or how you wanna call it).
* we need to map the vehicle to the route / itinerary (where do we put the reference? in the vehicle or in the itinerary?)
* about the mission, I am not that sure, we have already serviceProvided, won't be  enough extending allowed values?

AnderPijoan:
Hi,

Our suggested Itinerary entity is very similar to the Trip you are using (https://github.com/ftcardoso/dataModels/blob/PublicTransportation/PublicTransportation/ServiceNetwork/Trip/doc/spec.md), with the difference that we wouldn't need to define the stops as any other entity can serve as a stop. Doing so we can create a Trip traveling by different entity types.

There are waste trucks that collect from waste containers + local shops (no container, directly from the employee). If you need to create the Stop entities you would be duplicating information as the original entity is already in your model.

> Mission which will represent what's the mandate of a vehicle. A Mission will be composed by multiple MissionItem entities each one indicating the concerned asset, the service to be executed over the asset and the scheduled dates. An attribute will indicate the route to be followed by a Vehicle when performing such mission.

Mission and MissionItem are almost the same as Trip and Stop with the addition that as you can use any existing entity in MissionItem you do not need to duplicate locations in Stops.

However, we are not really sure about the Route entity:

> Route, agnostic about the vehicle which is going to be assigned to such route. A route will be represented by a MultiLineString in GeoJSON. A route can have a name, description, or even the service to which the route is related to.

If we start decomposing everything into list of "subentities", a route should be composed of multiple RoadSegments.

I would prefer the Trip approach but being able to use any entity inside its segments.

chicco785:
Hi,
My suggestion wasn't about replacing your proposal with Ubiwhere one, but merging the two to cover different needs and avoid potential duplication (and increase adoption thus!).

I will try to come with a proposal we can discuss if you think (maybe in gdoc), and involve also @ftcardoso. feelings?

Federico

chicco785:
Hi @AnderPijoan,
not sure you read my last comment, will you be fine in working on defining common models for the part of the scenario shared by Waste Management and Public Transport?

In general I think that the "shared" part could be:
- A model covering the itinerary plan and its schedule; (you would have few of this one per vehicle, maybe only one)
- A model covering the actual itinerary. (you would have endless "tracks" for each vehicle, that you can use also to analyse if you are "in line" with the itinerary plan)

Cheers,
Federico

AnderPijoan:
Yes, of course. I can try to create an Itinerary model suitable for being used both in Public Transportation and Waste Management. It would be a similar approach to what we have spoken here and disengaged from any specific domain.

AnderPijoan:
Good morning,

Created another datamodel for Transport Service Networks, by merging the comments provided here with the fork suggested (https://github.com/ftcardoso/dataModels/blob/PublicTransportation/PublicTransportation/ServiceNetwork/Trip/doc/spec.md). 

We have omitted some public transport domain specific attributes from the Itinerary entity (such as bikesAllowed, wheelchairAccessible, etc.) but we suppose that it will be straightforward for people that use GTFS to include them in their models. And we have moved the vehicle type from route to itinerary as some routes may be provided by different types of vehicles.

Entities are structured as follows:

+ Agency : The company/agency that provides the service.

+ Route : A group of itineraries that provide the same service:
+ A public transport line between two locations, for example : Route Avenida de América to Atocha which will have several frequencies (modeled as itineraries) along the day.
+ A waste collection service for a municipality, for example : Bilbao glass waste collection service, which will have several itineraries performed by the available collection trucks.
+ A delivery service for a logistic company, for example : Deliveries in Madrid municipality , which would contain several itineraries performed by different delivery vans.

+ Itinerary : Each itinerary that is performed to give the service of the route. Itineraries can be all the frequencies of a public transport route, all the trips travelled by the waste collection vehicles, all the trips travelled by the delivery vehicles, etc.

+ Itinerary segment : Inside an itinerary, several points are visited for interacting with some entity there: 
+ Interacting with a public transport stop by dropping and getting passengers
+ Interacting with a waste container to empty it
+ Interacting with a user needing a delivery etc.
For this purpose, each itinerary segment will have a reference to the start entity of the segment, end entity of the segment, scheduled paths, times and really performed paths and times.

Let me know your opinion. Best regards.

Transportation/ServiceNetwork/Itinerary/doc/spec.md
```diff
@@ -0,0 +1,88 @@
...
+    + Optional
+
++ `scheduledPath` : Itinerary's scheduled path geometry represented by a GeoJSON LineString. 
+    + Normative References: [https://schema.org/description](https://schema.org/description)
+    + Optional
+
++ `performedPath` : Performed path of the vehicle, to be updated while vehicle is travelling. A geometry represented by a GeoJSON LineString. 
+    + Normative References: [https://schema.org/description](https://schema.org/description)
+    + Optional
+
```
chicco785: wouldn't it be better to decouple the modelling of the planned itinerary from the one of performed path? 
having a separate entity to model the actual itinerary, will avoid you to duplicate every time the static information of the itinerary. for example if you have a planned itinerary for a vehicle during a year, the plan want change, the only thing that will change is the actual itinerary run every day by the vehicle.

Transportation/ServiceNetwork/ItinerarySegment/doc/spec.md
```diff
@@ -0,0 +1,93 @@
...
+
++ `arrivalPoint` : Departure location represented by a GeoJSON Point.
+    + Attribute type: `geo:json`.
+    + Normative References: [https://tools.ietf.org/html/rfc7946](https://tools.ietf.org/html/rfc7946)
+    + Mandatory if `refArrival is not present.
+
++ `refDeparture` : Segment's departure entity's id.
+    + Attribute type: [StructuredValue](https://schema.org/StructuredValue).
+        + Subproperties:
+            + `entityType` : Entity type which is the departuring point.
```
chicco785: i don't think you need the entityType and so a structuredValue the link to entity will be enough (then when you retrieve the entity, you anyhow have the information on the entity type. @jmcanterafonseca what do you think ?

Transportation/ServiceNetwork/ItinerarySegment/doc/spec.md
```diff
@@ -0,0 +1,93 @@
...
+                + Type: [Text](http://schema.org/Text)
+            + `id` : Entities id.
+                + Type: List of [Text](http://schema.org/Text)
+    + Optional
+
++ `refArrival` : Segment's arrival entity's id.
+    + Attribute type: [StructuredValue](https://schema.org/StructuredValue).
+        + Subproperties:
+            + `entityType` : Entity type which the arrival point.
```
chicco785: see comment above

Transportation/ServiceNetwork/ItinerarySegment/doc/spec.md
```diff
@@ -0,0 +1,93 @@
...
+    + Attribute type: [Time](http://schema.org/Time) or ```ISO8601``` (legacy).
+    + Mandatory
+
++ `arrivalTimestamp`:	Timestamp which captures when the user finished the trip segment. This value can also appear as a FIWARE [TimeInstant](https://github.com/telefonicaid/iotagent-node-lib/blob/develop/README.md#TimeInstant)
+    + Attribute type: [Time](http://schema.org/Time) or ```ISO8601``` (legacy).
+    + Optional
+
+## Example
+
+    {
```
chicco785: it would be better to extend the example to include all the modelled fields

Transportation/ServiceNetwork/Itinerary/doc/spec.md
```diff
@@ -0,0 +1,88 @@
...
++ `description` : Description about the itinerary. 
+    + Normative References: [https://schema.org/description](https://schema.org/description)
+    + Optional
+
++ `refRoute`: Route this itinerary belongs to.
+    + Attribute type: Reference to a [Route](../../Route/doc/spec.md)
+    + Mandatory
+
++ `vehicleType`: Describes the type of transportation used on a route
+    + Attribute type: [Number](https://github.com/schema.org/Number)
```
chicco785: according to the allowed values, it's not a number

Transportation/ServiceNetwork/Itinerary/doc/spec.md
```diff
@@ -0,0 +1,88 @@
...
+      "type": "Itinerary",
+      "name": "Itinerary 12",
+      "description": "Bus Itinerary for 2017-09-22",
+      "segeemnts": [
+            { "id": "wastecontainer:27" , "type": "WasteContainer" , "dateScheduled" : "2017-09-22T15:05:59.408Z" },
+            { "id": "wastecontainer:92" , "type": "WasteContainer" , "dateScheduled" : "2017-09-22T15:09:23.788Z" },
+            ],
+      "refAssignedVehicle" : "vehicle-512"
+    }
+
```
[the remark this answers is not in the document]
chicco785: in the model segments is not defined. i suppose that with the new model, either the Itinerary includes a list of references to IntinerarySegment or the IntinerarySegment should include a link to the Intinerary.

Transportation/ServiceNetwork/Itinerary/doc/spec.md
```diff
@@ -0,0 +1,88 @@
...
+        + `bus` : Used for short- and long-distance bus routes
+        + `ferry` : Used for short- and long-distance boat service
+        + `cablecar` : Used for street-level cable cars where the cable runs beneath the car
+        + `gondola` : Suspended cable car. Typically used for aerial cable cars where the car is suspended from the cable
+        + `funicular` : Any rail system designed for steep inclines
+        + `car` : Particular vehicle
+        + `van` : 
+        + `truck` : 
+        + `electriccar` :
+        - Any other value meaningfull for the scenario
```
jmcanterafonseca: redefining things already defined .... in Vehicle

Transportation/ServiceNetwork/Itinerary/doc/spec.md
```diff
@@ -0,0 +1,88 @@
...
+    + Optional
+
+## Example
+
+    {
+      "id": "itinerary:89237",
+      "type": "Itinerary",
+      "name": "Itinerary 12",
+      "description": "Bus Itinerary for 2017-09-22",
+      "segeemnts": [
```
jmcanterafonseca: typos

Transportation/ServiceNetwork/ItinerarySegment/doc/spec.md
```diff
@@ -0,0 +1,93 @@
...
++ `name` : Name given to the itinerary
+    + Normative References: [https://schema.org/name](https://schema.org/name)
+    + Optional
+
++ `description` : Description about the itinerary. 
+    + Normative References: [https://schema.org/description](https://schema.org/description)
+    + Optional
+
++ `scheduledPath` : Itinerary's scheduled path geometry represented by a GeoJSON LineString. 
+    + Normative References: [https://schema.org/description](https://schema.org/description)
```
jmcanterafonseca: what has to do with description?

Transportation/ServiceNetwork/Route/doc/spec.md
```diff
@@ -0,0 +1,48 @@
...
+	- Attribute type: https://schema.org/Text
+	- Optional 
+
+- ```refAgency```: Agency this route operated by.
+        - Attribute type: Reference to a [Agency](../../Agency/doc/spec.md)
+        - Mandatory
+
+- ```color```: Color that corresponds to the route. The color must be provided as a six-character hexadecimal number, for example, 00FFFF.
+	- Attribute type: [Text](https://schema.org/Text)
+	- Optional
```
jmcanterafonseca: is this REALLY an attribute of a route?

Transportation/ServiceNetwork/Route/doc/spec.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## Example
+
+```
+{
+	"id": "routeID",
+        "type": "Route",
+	"short_name": "SAUSA"
```
jmcanterafonseca: inconsistent attribute name

Transportation/ServiceNetwork/Route/doc/spec.md
```diff
@@ -0,0 +1,48 @@
...
+
+- ```longName```: Full name of a route, more descriptive than the ```route_short_name``` and often include the route's destination or stop
+    - Normative References: https://schema.org/name
+    - Mandatory 
+
+- ```description```: Description about this route.
+	- Attribute type: https://schema.org/Text
+	- Optional 
+
+- ```refAgency```: Agency this route operated by.
```
jmcanterafonseca: I would name this a `operatedBy`

Transportation/ServiceNetwork/Route/doc/spec.md
```diff
@@ -0,0 +1,48 @@
...
+## Data Model
+
+- ```id```: Entity's unique identifier.
+
+- ```type```: Entity type. It must be equal to ```Route```.
+
+- ```shortName```: Short name of a route, often a short and abstract identifier like "7", "34", or "Blue" that riders use to identify a route.
+    - Normative References: https://schema.org/name
+    - Mandatory 
+
```
jmcanterafonseca: please name this as just `name`

Transportation/ServiceNetwork/ItinerarySegment/doc/spec.md
jmcanterafonseca: yes, totally agreed, this is not the way of modelling things using the NGSI data model.

Transportation/ServiceNetwork/Itinerary/doc/spec.md
jmcanterafonseca: +100

Transportation/ServiceNetwork/Route/doc/spec.md
AnderPijoan: It is quire common when mapping Public Transport GTFS files which usually contain a color for each route. But it is something not really related to a route.

AnderPijoan:
Hi @jmcanterafonseca , regarding this request and the following proposals we have for the data model, we plan to attend the Fiware Summit at Malaga this month. 

I think some partners from our project tried to contact some people at the Summit but just in case I want to ask if it would be possible to arrange some meeting or even small talk about the model we are proposing for waste management and the issues you may see. We think it will be more useful and faster than writing here.

Let me know if there is someone we can talk to. Thank you.

chicco785:
hi @AnderPijoan i will be around, and surely i think we could try to call for a meeting on revising / proposing data models if the agenda allows.

AnderPijoan:
@chicco785 Is it possible to arrange and book a room for the meeting from now? In order to find a a time when all of you can.

jmcanterafonseca:
I would meet just informally around no need to set up a meeting.

just to be clear we are looking for datamodels that:

A/ are generic enough to be applied to many different applications, not a
single one, because otherwise the data model becomes too specific and there
is no point in publishing it.
B/ Are well aligned with the principles and phylosophy of NGSI (attributes,
metadata), in general compound structures are not a very good idea
C/ From Complex data models or application specific can be extracted pieces
that can be generic enough to be adopted ...

2017-11-09 9:41 GMT+01:00 Ander <notifications@github.com>:

AnderPijoan:
Yes, thats it. 

However there are about 20 models that we want to discuss. So it would take us some time. As part of our project we need to have this sorted as soon as possible so sorry for being insistent but I dont want to happen that in the end the agenda does not allow us to take some time to get together.

So thats why I would like if possible to set some approximate day and time for the meeting where you have the agenda free.

Best regards.

jmcanterafonseca:
28th from 17:30 to 18:00 can work for me.

If you have 20 models have you considered to publish them as a
complementary data models project focused on a particular domain?
There could be a link between the curated Fiware Data Models project and
your project. We don't have bandwidth nor domain experience to review them
so ... I think this could be a good tradeoff



2017-11-23 10:10 GMT+01:00 Ander <notifications@github.com>:

AnderPijoan:
Some of them are extensions of the existing models at Fiware Schema, others could be useful for adding them to the schema and other are too domain related and have no sense to be added:

## Waste classification (extension of WasteManagement [1]):

- Waste : definition of the formerly called Litter [1]
- Waste Category : definition of the formerly called LitterCategory [1]
- Sorting type : New model for group of wastes that are collected in a legislation.

[1] http://fiware-datamodels.readthedocs.io/en/latest/WasteManagement/doc/introduction/index.html

## Waste management (extension of WasteManagement [1]):

- CollectionPoint : turn and extend the formerly called WasteContainer [2] to a wider one that suits all waste management systems (containerized, bags, door2door, etc.)
- CollectionPointType : the same with WasteContainerModel [3]
- CollectionPointIsle : the same with WasteContainerIsle [4]

[2] http://fiware-datamodels.readthedocs.io/en/latest/WasteManagement/WasteContainer/doc/spec/index.html
[3] http://fiware-datamodels.readthedocs.io/en/latest/WasteManagement/WasteContainerModel/doc/spec/index.html
[4] http://fiware-datamodels.readthedocs.io/en/latest/WasteManagement/WasteContainerIsle/doc/spec/index.html 

## Transportation (new models suitable for being added to transportation [5]):

- Agency : new model for itinerary agencies as discussed in this pull request.
- Route : new model for a route that contains several itineraries for different datetimes.
- Itinerary : Each itinerary belonging to a route
- ItinerarySegment : Each step of an itinerary.

[5] http://fiware-datamodels.readthedocs.io/en/latest/Transportation/doc/introduction/index.html

## KeyPerformanceIndicator (extension of the entity [6]):

- KeyPerformanceIndicator : extension of the model attributes [6] to contain more KPI information.

[6] http://fiware-datamodels.readthedocs.io/en/latest/KeyPerformanceIndicator/doc/spec/index.html

## Strategies (suitable for being added?):

- Strategy : A plan carried out in time and place to get some effect in a target group.
- Action : Actions that are carried out in a strategy.

------

The rest of the model we will be using is very domain specific so we won't be talking about it.

albertoabellagarcia:
I am Alberto Abella, the responsible person for curating data models. I'd like to comment on this PR.
1) this repository is frozen, and therefore this PR has to be moved the new one here (https://github.com/smart-data-models/)
2) It is also true that the documentation structure is not the same and it has to be adapted.
3) Every change in a single data model has to be in a single PR for simpler analysis
you can contact me at alberto . abella @ fiware. org for further comments on this

PR #146, opened 2017-10-03
Example Bike hire add

jmcanterafonseca:


jmcanterafonseca:
@sergg75 a review is needed. 

On the other hand shouldn't the `status` attribute be just a string instead of an array of strings? If so, I will make the change in this PR as well

Issue #127, opened 2017-05-12
refactor point of interaction and fix file location


jmcanterafonseca:

jmcanterafonseca:
@FRDavid please could you work to fix this issue?

FRDavid:
I'm sorry @jmcanterafonseca, I missed the first message and I'm out of context,

What should I do to complete the task? Could you specify more?

It's about create the [harmonised definition](https://github.com/GSMADeveloper/HarmonisedEntityDefinitions/blob/master/PointOfInterest.md)?

It's about file organization?
- doc/spec.md
- example.json
- schema.json

Best regards and sorry for the inconvenience

jmcanterafonseca:
yes  you need to have the same file organization as other data models. That includes creating a schema.json and example.json consistent with the specification of the data model.

FRDavid:
Ok in file organization. What do you mean with refactor?

jmcanterafonseca:
for the time being nothing ... but I don't recall

2017-10-03 13:10 GMT+02:00 David <notifications@github.com>:

jmcanterafonseca:
fixed

PR #148, opened 2017-10-03
Move example.json and schema.json out of the doc folder

FRDavid:
Hi @jmcanterafonseca , I moved the example.json and schema.json to dispose the files in a similar way to other data models. Please, check if this is the desired change by #127.

Let me know if more changes are required.

BTW, in a short future I will propose a new data model, the idea is provide information about detected near wifi devices.


jmcanterafonseca:
LGTM

thanks

Issue #150, opened 2017-10-03
check validity of examples against each json schema


jmcanterafonseca:
it is needed to add to the travis build the check of the validity of each example.json against each JSON Schema. please check the validate.sh in order to find how to check the validity of JSON data against JSON Schemas using the ajv tool

jmcanterafonseca:
@cclauss would you like to work on this?

cclauss:
We would need to add the [ajv](https://github.com/epoberezkin/ajv) tool to the repo.  It is not clear to me how to do that.  Once it is added to the repo, I could work on mods to validate.sh

chicco785:
hi, i need to review it, but i create a node-js script that uses ajv to parse all the folders to provide results. this could be called from the .travis.yml to validate pull requests.

chicco785:
I created a pull request with the validator #160

jmcanterafonseca:
closing as this has been fixed with the landing of the validator

PR #160, opened 2017-10-12
DataModel validator

chicco785:
FIWARE Data Model validator is an utility to help the management of NGSI DataModels. Its code leverage on the [AJV JSON Schema Validator](https://github.com/epoberezkin/ajv).

The FIWARE Data Model validator perform the following checks for each Data Model:
* validity of JSON Schema
* validity of JSON Examples
* adherence of Data Model name to FIWARE Data Models guidelines
* existence of Doc folder
* existence of JSON Schema
* existence of one or more JSON examples
* existence of README file

This function could be used to support the continuous integration life cycle of fiware data models. see #150.


cclauss:
Cool!  The next step will be to change [this line in .travis.yml](https://github.com/Fiware/dataModels/blob/master/.travis.yml#L14) to run the code in this PR and/or  _validate.sh_ on every new pull request.  How do we make that happen?

chicco785:
I have to check how in the .travis.yml we can include nodejs beside python support. I know how to make it work in .travis.yml that use nodejs as language... i would do some test and commit the extended .travis.yml.

chicco785:
@cclauss not sure its the best way of doing a .travis.yml with multiple languages, but it's the only one i manage to make work. by the way, one example does not validate correctly:

Error: {
	"Transportation": [
		"Transportation/Bike/BikeHireDockingStation: Example Transportation/Bike/BikeHireDockingStation/example.json is invalid: [{\"keyword\":\"type\",\"dataPath\":\".status\",\"schemaPath\":\"#/allOf/2/properties/status/type\",\"params\":{\"type\":\"array\"},\"message\":\"should be array\"}]"
	]
}

chicco785:
in fact the value of status should be an array according to the schema:
"status": ["working"]

chicco785:
@jmcanterafonseca let me know when the pull request #164 to fix the example is approved, so that I merge the new master with this code and pass the travis-ci

cclauss:
@chicco785 This is really cool stuff that you are doing.  I am trying to track it all, given that I don't know node.  Instead of changing the language from Python to Node in .travis.yml, would it be possible to take all of those steps that you are adding to .travis.yml and instead put those steps into __validate.sh__.

Then in Travis, we could then run __validate.sh__ which would:
1. install node_js
2. npm install -g fiware-model-validator
3. validate -p .

The reason that I would recommend this approach is because flake8 will generate different results on Python 2 and on Python 3 and compatibility with both is important at least until 2020.  Making node the Travis language has the effect of having all flake8 tests run in only one version of Python.    Travis used to do two test runs in parallel but this PR unfortunately removes that.

Also, we do not need sudo on pip in a virutalenv because, by definition, we are the only ones in that environment.

Finally, if the above is not possible or not desirable, it might also be possible to rely on the node 7 that is built into the Travis environment even when Python is the language https://travis-ci.org/Fiware/dataModels/builds/286899391#L392-L397 and then just do steps 2 and 3 in the Travis file.

Your thoughts on how to get the best of both worlds?

chicco785:
Hi,
Sure, I will try as you mentioned, I thought I could exploit the matrix construct, but that didn't work (maybe is my fault), and the annoying thing is that I don't have a local travis environment to test, so I stopped at the first successful configuration. (Which I knew was not the best one).

Do you know if it is possible to run travis locally? I looked around but didn't find any good solution.



iFede - Fede's mobile edition

cclauss:
I do not know how to run Travis locally, but I do know how to set up my own repos that git clone in the source code from other repos and test them...  See https://github.com/cclauss/itinerant-tester  It is not local but at least it is not in the repo being tested.

validator/validate.js
```diff
@@ -0,0 +1,482 @@
...
+      validate = ajv.compile(schema);
+      /* istanbul ignore else */
+      if (typeof validate == 'function') {
+        addValidSchema(fullPath, 'Schema '+ file +' is valid');
+      } else {
+        addError(fullPath, 'Schema '+ file +' failed to compile to a function');
+        if(failErrors) throw new Error(validate.errors);
+      }
+    } else {
+      /*
```
jmcanterafonseca: code commented

chicco785: removed

chicco785:
now i am adding node on top of python as suggested by @cclauss, still given that bike schema is incorrect (see pull request #164), of course the ci process will fail.

cclauss:
@chicco785 This is an awesome piece of work!  Congratulations.

chicco785:
many improvements are possible :) such as supporting the asynch compile of schemas, let's see when me or someone else in my time has some cycle for that.

.travis.yml
```diff
@@ -1,17 +1,22 @@
...
   # exit-zero treats all errors as warnings.  GitHub editor is 127 chars wide
   - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 script:
   - true  # add other tests here
+  - validate -p .
+
 notifications:
   on_success: change
-  on_failure: always
+  on_failure: always
```
jmcanterafonseca: new line at the end of file needed

validator/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Using the validator
+To use the validator, execute it from the root of the DataModel repository:
+```
+validate -p DataModel -w ignore -i [common-schema.json,geometry-schema.json]
+```
+
+Command line available options are:
+
+*  ``-i, --fiware:importSchemas``.  Additional schemas that will be included during validation. Default imported schemas are: common-schema.json, geometry-schema.json   [array]
```
jmcanterafonseca: do we need to prepend the word `fiware` to all the options?

validator/config.json.example
```diff
@@ -0,0 +1,20 @@
+{
+  "fiware": {
```
jmcanterafonseca: I would name this section as  `dataModel` instead of `fiware`

validator/package.json
```diff
@@ -0,0 +1,31 @@
+{
+	"name": "fiware-model-validator",
+	"version": "0.1.4",
+	"description": "A simple validator for FIWARE Data Models",
+	"keywords": [
+		"node",
+		"fiware"
```
jmcanterafonseca: please add more meaningful keywords

validator/validate.js
```diff
@@ -0,0 +1,470 @@
+#!/usr/bin/env node
+
+var fs = require("fs");
```
jmcanterafonseca: `var` missing in the rest of declarations

validator/validate.js
```diff
@@ -0,0 +1,470 @@
+#!/usr/bin/env node
+
```
jmcanterafonseca: "use strict";

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+if (nconf.get('ajv:missingRefs') == null) {
+  nconf.set('ajv:missingRefs', 'true');
+}
+if (nconf.get('ajv:extendRefs') == null) {
+  nconf.set('ajv:extendRefs', 'fail');
+}
+if (nconf.get('ajv:allErrors') == null) {
+  nconf.set('ajv:allErrors', true);
+}
+
```
jmcanterafonseca: this block of code could be resolved more elegantly by iterating over a hash

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+  if (!stats.isDirectory()) {
+    throw new Error("The path passed must be a directory");
+  }
+}
+catch (err) {
+  process.exitCode = -1;
+  console.error("\n Invalid Path: "+err.message+"\n");
+  return;
+}
+
```
jmcanterafonseca: too long module. this would need to be properly restructured into Node modules and have a clear entry point

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+var ignoreFolders = nconf.get('fiware:ignoreFolders');
+var docFolders = nconf.get('fiware:docFolders');
+ignoreFolders = ignoreFolders.concat(['.git','node_modules']);
+var warningChecks = nconf.get('fiware:warningChecks');
+var externalSchemaFolders = nconf.get('fiware:externalSchemaFolders');
+var ignoreWarnings = (nconf.get('fiware:warnings') == 'ignore');
+var failWarnings = (nconf.get('fiware:warnings') == 'fail');
+var failErrors = !nconf.get('ajv:allErrors');
+
```
jmcanterafonseca: you could save many variable declarations by using a hash

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+    process.exit(2);
+  }
+  return json;
+};
+
+//check if a documentation file exists in a given path
+var docExist = function (fullPath){
+  if (!fileExists(fullPath, "spec.md") && !fileExists(fullPath, "introduction.md"))
+  if (addWarning(fullPath, "does not include a documentation file name spec.md or introduction.md") && failWarnings)
+    throw new Error("Fail on Warnings: " +JSON.stringify(warnings,null, '\t'));
```
jmcanterafonseca: missing space. please run a linter against your code JSHint + fixjsstyle could do the trick

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+//add valid schema to the valid schemas map for a given model
+var addValidSchema = function (modelPath, message) {
+  return addMessageToMap(modelPath, message,validSchemas);
+};
+
+//add valid example to the valid examples map for a given model
+var addValidExample = function (modelPath, message) {
+  return addMessageToMap(modelPath, message,validExamples);
+};
+
```
jmcanterafonseca: refactor your code in modules

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+var addUniqueToArray = function (array1, array2){
+  var result =  Array.from(array1);
+  array2.forEach(function(item2){
+    if (!array1.includes(item2))
+    result.push(item2);
+  });
+  return result;
+}
+
```
jmcanterafonseca: refactor, add this to an auxiliary module

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+    try {
+      var fullPath;
+      if (basePath != ".") fullPath = basePath + path.sep + fileName;
+      else fullPath = fileName;
+      var stat=fs.lstatSync(fullPath);
+
+      if (stat && stat.isDirectory()) {
+        // Dive into the directory
+        if (!ignoreFolders.includes(path.basename(fullPath)) && !docFolders.includes(path.basename(fullPath)) && !externalSchemaFolders.includes(path.basename(fullPath))){
+          if (warningChecks.includes("modelNameValid") && !ignoreWarnings) modelNameValid(fullPath);
```
jmcanterafonseca: perhaps ignoreWarnings can only need to be checked one time

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+  });
+};
+
+console.log("*** Active Warnings ***:" +warningChecks);
+dive(nconf.get('fiware:path'),commonSchemas);
+console.log("*** ValidSchemas ***: "+JSON.stringify(validSchemas,null, '\t'));
+console.log("*** ValidExamples ***: "+JSON.stringify(validExamples,null, '\t'));
+console.log("*** Warnings ***: "+JSON.stringify(warnings,null, '\t'));
+console.log("*** Errors ***: "+JSON.stringify(errors,null, '\t'));
+if (Object.keys(errors).length != 0) throw new Error(JSON.stringify(errors,null, '\t'));
```
jmcanterafonseca: refactor and define a clear modular structure

validator/validate.js
```diff
@@ -0,0 +1,470 @@
...
+};
+
+//load  a list of files, supports patterns, e.g. *-schema.json
+function getFiles(args) {
+  var files = [];
+  if (Array.isArray(args)) args.forEach(_getFiles);
+  else _getFiles(args);
+  return files;
+
+  function _getFiles(fileOrPattern) {
```
jmcanterafonseca: not a good idea, use modules

validator/validate.js
chicco785: it's a bug fix, due to the fact that the nconf.default of the library (nconf being a sort of hash), is not working properly.

chicco785:
@jmcanterafonseca the refactor will take a while (i have few deadlines to follow up), let me know if you prefer to just get fixed major issues and merge this initial version, or postpone to the overall refactoring

validator/config.json.example
chicco785: i used dmv for brevity

validator/validate.js
chicco785: fixed

validator/README.md
chicco785: replaced with dmv

jmcanterafonseca:
@chicco785 I think it would be a good idea to fix the most urgent refactoring issues now and then later do a more extensive refactoring work.

validator/validate.js
jmcanterafonseca: then you can move this code out to another module so that it can be properly removed when the bug from nconf is fixed. Another option would be to overwrite the nconf functions, all in all this is just Javascript

thanks!

.travis.yml
```diff
@@ -1,17 +1,22 @@
 language: python
 cache: pip
+node_js:
+  - "7"
```
cclauss: I think you need to remove these two lines before we merge or we turn this Travis file into a node_js tester instead of a Python tester and we would lose the Python2 vs. Python3 tests.

Also, you might want to include the linting changes from https://github.com/Fiware/dataModels/pull/176/files and then we could just drop #176 ...  Your call.

cclauss:
Here is an example of a Travis test that mixes two different testing languages https://travis-ci.org/getsentry/sentry/builds/291509879?utm_source=github_status&utm_medium=notification

chicco785:
hi @cclauss python v2 vs python v3 are working correctly (simply there is not nodejs versioning), anyhow the matrix example you linked is quite interesting and we should try to have a similar approach.

currently, while the code should be ok (when i run locally everything works), i am having an issue on the validation. it seems that for whatever reason some json files get truncated :S

chicco785:
- a major refactoring was applied trying to fix most of the request changes by @jmcanterafonseca, still an additional clean up may be good
- #180 fix a bug that will allow again the datamodel to compile correctly
- .travis.yml now use matrix construct suggested by @cclauss

chicco785:
also the linting is included, but as @cclauss mentioned, the old js code within the models need cleans up to pass the linting

chicco785:
@cclauss see the build split with the new .travis.yml configuration: https://travis-ci.org/Fiware/dataModels/builds/291561539?utm_source=github_status&utm_medium=notification

jmcanterafonseca:
@chicco785  this PR needs more work particularly on the JS linting side, please use fixjsstyle and ensure everything is properly indented, lines of code not too long, etc.

.travis.yml
```diff
@@ -18,6 +18,8 @@ matrix:
   ## linters for json and JavaScript files
     - language: node_js
       python:
+      node_js:
+        - 7
```
chicco785: @cclauss we need to have node versioning, if not travis-ci will install 0.10 version, see the build after your PR #184
https://travis-ci.org/Fiware/dataModels/jobs/292149216

cclauss:
Let' me test my understanding... In the node_js Travis pass, we are going to:
1. eslint --ext .json .
2. jshint .
3. validate -p .
Do I have that right?

chicco785:
yes, but for informative purposes, i would keep the linting and the
validation in tow different builds (but this is my personal preference, and
not a strong one)

.travis.yml
```diff
@@ -8,10 +8,24 @@ install:
...
+matrix:
+  include:
+    - language: node_js
+      python:
+      node_js:
+        - 7
+      install:
+        - npm install -g fiware-model-validator
+        - npm install -g jshint
+      before_script:
```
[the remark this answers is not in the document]
chicco785: #185

cclauss:
My sense if they are long runs then it is worth the complexity to separate into parallel passes but if they all run back-to-back in a single pass in a few minutes (like the Python runs) then it will be less confusing to consolidate the three steps above in a single pass.

validator/libs/checks.js
```diff
@@ -0,0 +1,110 @@
+/* Data Model Validation checkers for structure */
```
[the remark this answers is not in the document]
chicco785: fixed

chicco785:
@cclauss has you can see from code, i merged the the validation and the linting. the draw back, is that at this moment, validation is not even executed because of the validation failure.

jmcanterafonseca:
Looks good enough for me, and I did some smoke testing, so landing! 

thanks Fede!

Issue #163, opened 2017-10-12
Example of BikeHireDockingStation is not valid according to the schema


chicco785:
https://github.com/Fiware/dataModels/blob/fb59108062c0e60bed2b039428e00a180bc0cea6/Transportation/Bike/BikeHireDockingStation/example.json#L18

"status": ["working"]

jmcanterafonseca:
schema should be changed, not the example, see my comment

PR #164, opened 2017-10-12
Fix BikeHireDockingStation example.json

chicco785:
as reported in #163


jmcanterafonseca:
status should be only a single value. I don't know if the inconsistency is in the schema but the example is correct 

On the other hand your PR is not clean, please could you include a PR with just the commits needed? and not the merge with master, etc. ? thanks!

chicco785:
fixed as by your comment

jmcanterafonseca:
LGTM, thanks

PR #166, opened 2017-10-12
Add owner to gsma common schema

chicco785:
Checking the original GSMA Device Data Model we found out that the owner attribute is missing. We believe this is an relevant attribute in several scenarios.
* Added owner property as array of uris to common-schema.json to match GSMA Device Data Model
* Added owner property to GSMA Device Data Model spec

### DIFF
--- a/Device/Device/doc/spec.md
+++ b/Device/Device/doc/spec.md
@@ -153,6 +153,10 @@ Obviously, in order to toggle the referred switch, this attribute value will hav
     + Attribute type: [DateTime](https://schema.org/DateTime)
     + Optional    
 
++ `owner` : Entity's owners: an array of entities that represents the owners of an Thing.
+    + Attribute type: List of [Person]( http://schema.org/Person) or [Organization](https://schema.org/Organization).
+    + Optional
+
 ## Examples
 
     {
@@ -170,6 +174,7 @@ Obviously, in order to toggle the referred switch, this attribute value will hav
       "value": "l=0.22;t=21.2",
       "deviceState": "ok",
       "dateFirstUsed": "2014-09-11",
+      "owner": ["http://person.org/leon"]
     }
 
 
--- a/Device/Device/doc/spec.md
+++ b/Device/Device/doc/spec.md
@@ -153,7 +153,7 @@ Obviously, in order to toggle the referred switch, this attribute value will hav
     + Attribute type: [DateTime](https://schema.org/DateTime)
     + Optional    
 
-+ `owner` : Entity's owners: an array of entities that represents the owners of an Thing.
++ `owner` : Entity's owners: an array of entities that represents the owners of a Device.
     + Attribute type: List of [Person]( http://schema.org/Person) or [Organization](https://schema.org/Organization).
     + Optional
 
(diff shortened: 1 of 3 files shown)


jmcanterafonseca:
the PR looks good but I don't like the spurious changes it is introducing, please could you check your editor settings?

chicco785:
the changes should be clean now. issue is with the abnormal amount of white spaces in several of the datamodel files that atoms automatically removes

Device/Device/doc/spec.md
```diff
@@ -153,6 +153,10 @@ Obviously, in order to toggle the referred switch, this attribute value will hav
     + Attribute type: [DateTime](https://schema.org/DateTime)
     + Optional    
 
++ `owner` : Entity's owners: an array of entities that represents the owners of an Thing.
```
jmcanterafonseca: The owner of a thing or the owner of the device?

common-schema.json
```diff
@@ -37,6 +37,12 @@
         },
         "dataProvider": {
           "type": "string"
+        },
+        "owner": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "format": "uri"
```
jmcanterafonseca: it is not necessary for it to be a URI, it could be a URI or an id (reference) pointing to another entity stored in the same server

Device/Device/doc/spec.md
chicco785: in general of an object (for the common schema), for the specific data model of a device

common-schema.json
chicco785: i added support for both using oneOf construct. alternatively i can simply remove the formatting specification.

common-schema.json
```diff
@@ -37,6 +37,21 @@
...
+          "items": {
+            "type": "string",
+            "oneOf": [ 
+              {
+                "format": "uri"
+              },
+              {
+                "minLength": 1,
+                "maxLength": 256,
+                "pattern": "^[\\w\\-\\.\\{\\}\\$\\+\\*\\[\\]`|~^@!,:\\\\]+$"
```
jmcanterafonseca: instead of repeating here the pattern you could just cross-reference it

 "$ref": "https://fiware.github.io/dataModels/common-schema.json#/definitions/EntityIdentifierType"

chicco785: indeed!

datamodel_template.md
```diff
@@ -20,6 +20,10 @@ A JSON Schema corresponding to this data model can be found {{add link to JSON S
     + Attribute type: [DateTime](https://schema.org/DateTime)
     + Optional    
 
++ `owner` : Entity's owners: an array of entities that represents the owners of an Thing.
```
jmcanterafonseca: a Thing

Device/Device/doc/spec.md
```diff
@@ -153,6 +153,10 @@ Obviously, in order to toggle the referred switch, this attribute value will hav
     + Attribute type: [DateTime](https://schema.org/DateTime)
     + Optional    
 
++ `owner` : Entity's owners: an array of entities that represents the owners of a Device.
+    + Attribute type: List of [Person]( http://schema.org/Person) or [Organization](https://schema.org/Organization).
```
jmcanterafonseca: list of references to, because we don't expect to inline them, do we?

chicco785: (y)

jmcanterafonseca:
LGTM, thanks

PR #172, opened 2017-10-17
Fix url string format PoinOfInteraction

chicco785:
afaik, url is not a valid format in json schema, replaced with uri
which is a super set of url, better would be probably to use a pattern.
feedbacks?


jmcanterafonseca:
LGTM thanks

PR #174, opened 2017-10-18
Remove anachronistic import with_statement

### LINKED ISSUE #169: What versions of Python are supported?
My assumption is all the currently supported versions of CPython: 2.7, 3.4, 3.5, and 3.6.  If so, we can safely remove the import of the with_statement which was only needed for Python 2.5 and earlier.
* https://github.com/Fiware/dataModels/search?utf8=✓&q=with_statement&type=

cclauss:
Fixes #169  Remove `from __future__ import with_statement` which has not been needed post Python 2.5.  Also, clean up flake8 issues in each file...
* Remove unused imports
* Remove or mark with _noqa_ unused variables
* Remove backslash line termination with parens or str.join()
* Use += instead of repeating variable name
* list comprehension, dict comprehension, dict.get(), str.format(), ternary if

### DIFF
--- a/Environment/AirQualityObserved/harvest/barcelona_airquality_harvest.py
+++ b/Environment/AirQualityObserved/harvest/barcelona_airquality_harvest.py
@@ -174,8 +174,9 @@ def get_air_quality_barcelona(target_stations):
 
                         # Entity id corresponds to the observed date starting
                         # period (in local time)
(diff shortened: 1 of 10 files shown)


jmcanterafonseca:
this PR scares me a lot. let's concentrate code changes on one issue at a time, so please stick to the  initial plan, which was removing the import with_statement ... 

thanks for the effort

PR #175, opened 2017-10-20
Remove anachronistic import with_statement (not so scary version)

### LINKED ISSUE #169: What versions of Python are supported?

cclauss:
Fixes #169 Remove `from __future__ import with_statement` which has not been needed post Python 2.5.


jmcanterafonseca:
LGTM thanks

PR #177, opened 2017-10-23
Use list comprehension to simplify data persistance

cclauss:


Transportation/Bike/BikeHireDockingStation/harvest/bicycle_hire_station_barcelona_harvest.py
```diff
@@ -149,22 +149,11 @@ def persist_data(entity_list):
 # Main module
 def main():
     data = read_data(SOURCE)
-
-    if data is None:
+    if data:
```
jmcanterafonseca: same comment, I prefer the existing loop rather than a magic one line sentence, which sometimes turned into something not very flexible

PR #179, opened 2017-10-23
Remove unused import, Simplify w/ dict comprehension

cclauss:
* str.join()
* ternary if
* suppress flake8 F841 unused variable warning
* use += instead of repeating variable name

### DIFF
--- a/Weather/WeatherForecast/harvest/portugal_weather_forecast_harvest.py
+++ b/Weather/WeatherForecast/harvest/portugal_weather_forecast_harvest.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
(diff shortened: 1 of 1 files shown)


Weather/WeatherForecast/harvest/portugal_weather_forecast_harvest.py
```diff
@@ -70,14 +70,8 @@ def decode_wind_direction(direction):
...
 def get_weather_forecasted():
-    data = {}
-
-    for locality in iptma_codes:
-        code = iptma_codes[locality]
-
-        data[code] = get_weather_forecasted_pt(locality)
-
-    return data
+    return {iptma_codes[locality]: get_weather_forecasted_pt(locality)
```
jmcanterafonseca: even though I appreciate the effort to make it more Pythonic, I think the original code is more evident for someone who is not a great Python expert

PR #180, opened 2017-10-23
fix missing } in common-schema.json

chicco785:
fix missing } in common-schema.json

PR #181, opened 2017-10-23
add-linter-for-json

chicco785:
this ensures that json files are valid json (which is a pre-requisite
to json schema validation).


chicco785:
this test correctly spot and provide feedback on issues like the one fixed in #180

PR #182, opened 2017-10-24
Refine a little bit the waste container model example

jmcanterafonseca:


jmcanterafonseca:
@chicco785 please could you r?

WasteManagement/WasteContainerModel/example.json
```diff
@@ -5,8 +5,8 @@
   "height": 0.8,
   "depth": 0.4,
   "cargoVolume": 150,
-  "brandName": "Contenedores Ejemplo",
-  "name": "C1",
+  "brandName": "Brute",
+  "name": "Dumpster_Brute_2009_Plastic_Green",
```
chicco785: I think it would be beneficial to add your explanation on the difference between the two in the spec.md, the example looks good.

jmcanterafonseca: fixed

Issue #151, opened 2017-10-03
provide missing JSON Schemas 


jmcanterafonseca:
there are some data models which JSON Schema is missing. We need to provide it together with the example.json

@cclauss would you like to work on this?

chicco785:
@jmcanterafonseca i have already worked out some of them... may be i can do few pull requests. but first I need to check what i did :) (was few months ago...)

jmcanterafonseca:
@chicco785 please could you make a pull request with the data models you may have and then we can assign someone to complete them?

jmcanterafonseca:
we currently have a nice pull request from XapiX that provides all the missing JSON Schemas. It will be landing soon ... ;)

jmcanterafonseca:
I think we can now close this

PR #191, opened 2017-10-30
Device datamodel json schema

chicco785:
Schemas and Examples for:
* Device
* DeviceModel

### DIFF
--- a/Device/Device/schema.json
+++ b/Device/Device/schema.json
@@ -54,28 +54,6 @@
             }
           ]
         },
-        "supportedProtocol": {
-          "type": "array",
-          "items": {
-            "type": "string",
-            "enum": [
-              "ul20",
-              "mqtt",
-              "lwm2m",
-              "http",
-              "websocket",
-              "onem2m",
-              "sigfox",
-              "lora",
-              "nb-iot",
-              "ec-gsm-iot",
-              "lte-m",
-              "cat-m",
-              "3g",
-              "grps"
-            ]
-          }
-        },
         "configuration": {
           "type": "object"
         },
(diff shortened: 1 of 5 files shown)


chicco785:
related to issue #151

Device/DeviceModel/schema.json
```diff
@@ -0,0 +1,165 @@
...
+              "metering",
+              "eventNotification"
+            ]
+          }
+        },
+        "supportedProtocol": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
```
jmcanterafonseca: instead of repeating these definitions (which are also at device level) I think we need to create one schema for definitions and reuse it as we did in other cases

chicco785: k

chicco785:
@jmcanterafonseca i created a common part of the schema (and rebased following your latest integrations of PR)

Device/device-schema.json
```diff
@@ -0,0 +1,79 @@
...
+      "type": "object",
+      "category": {
+        "$ref": "/definitions/CategoryType"
+      },
+      "controlledProperty": {
+        "$ref": "/definitions/ControlledPropertyType"
+      }
+    }
+  }
+}
```
jmcanterafonseca: nit: new line needed

Device/Device/schema.json
```diff
@@ -0,0 +1,152 @@
...
+            {
+              "format": "ipv6"
+            }
+          ]
+        },
+        "supportedProtocol": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
```
jmcanterafonseca: we need this enumeration to be part of the commons as well

chicco785: good spot, i missed that, new commit is incoming

PR #204, opened 2017-12-19
Add illuminance to the weather observed information model

jmcanterafonseca:


jmcanterafonseca:
@chicco785 r?

chicco785:
hi @jmcanterafonseca looks good to me, i am only not sure illuminance is part of weatherObservation.
like we have airquality, it could make sense to have a model specific to light pollution

jmcanterafonseca:
hi, illuminance has something to do with weather observation, for instance low illuminance might mean overcast weather.

Issue #223, opened 2018-02-26
Minor glitches in the description of the location and address attributes of the BikeHireDockingStation spec


aarranz:
Currently, the [spec document](https://github.com/Fiware/dataModels/blob/079dca98e6122cc9105bc4a7b36c315cf0d763da/Transportation/Bike/BikeHireDockingStation/doc/spec.md) describes the `address` attribute as to be "Mandatory if `location` is not defined". This description is strange because the `location` attribute is described to be mandatory, so it will be clearer to mark the `address` attribute directly as optional.

Another option would be to mark the `location` attribute "Mandatory if `address` is not defined." as we do on other specs.

jmcanterafonseca:
@aarranz I would prefer to mark location as "Mandatory if address is not present"

jmcanterafonseca:
@aarranz please could you submit a PR fixing this?

aarranz:
The PR was merged, so I think this issue can be closed

PR #225, opened 2018-02-26
BikeHireDockingStation Spec: Mark location as mandatory if address is not provided

aarranz:
This pull request fix issue #223

### DIFF
--- a/Transportation/Bike/BikeHireDockingStation/doc/spec.md
+++ b/Transportation/Bike/BikeHireDockingStation/doc/spec.md
@@ -8,7 +8,7 @@ features and availability of bikes and free slots.
 ## Data Model
 
 
-A JSON Schema corresponding to this data model can be found [here](http://fiware.github.io/dataModels/Transportation/BikeHire/schema.json).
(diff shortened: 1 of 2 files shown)


jmcanterafonseca:
@aarranz we need to update the JSON Schema as well. also the link to the JSON Schema is broken in the spec page, so a good opportunity to fix it as well. 

Thanks!

aarranz:
I have added a commit fixing the link on the documentation. The schema was fixed on `af2a05b`, please review if that change is enough or if you require more changes.

Transportation/Bike/BikeHireDockingStation/doc/spec.md
```diff
@@ -8,7 +8,7 @@ features and availability of bikes and free slots.
 ## Data Model
 
 
-A JSON Schema corresponding to this data model can be found [here](http://fiware.github.io/dataModels/Transportation/BikeHire/schema.json).
+A JSON Schema corresponding to this data model can be found [here](../schema.json).
```
jmcanterafonseca: we want a link to the raw content not to the resource, so you need to keep the URL but pointing to the proper resource. thanks!

Issue #221, opened 2018-02-20
modeling pollen level to measure air quality


chicco785:
we need to model level of pollens in the air that may affect the life quality of people.

we thought, basically of two options:

* reuse the AirQualityObserved model, by using instead of the chemical pollutant
the name of the allergene for example (the problem here would be that we need to add new values to the level values):

    ```
    {
      "id": "AirQualityObserved-CDMX-Pollen-Cuajimalpa",
      "type": "AirQualityObserved",
      "Alnus_Level": {
        "type": "Text",
        "value": "moderate",
        "metadata": {
          "description": {
            "type": "Text",
            "value": "Concentration level of gp per m3 - Allergenicity: A "
          },
          "referenceSpecification": {
            "type": "Text",
            "value": "https://www.uco.es/rea/infor_rea/interpretacion.html"
          }
        }
      },
      "Casuarina_Level": {
        "type": "Text",
        "value": "low",
        "metadata": {
          "description": {
            "type": "Text",
            "value": "Concentration level of gp per m3 - Allergenicity: B "
          },
          "referenceSpecification": {
            "type": "Text",
            "value": "https://www.uco.es/rea/infor_rea/interpretacion.html"
          }
        }
      },
      "Cupressaceae_Level": {
        "type": "Text",
        "value": "moderate",
        "metadata": {
          "description": {
            "type": "Text",
            "value": "Concentration level of gp per m3 - Allergenicity: A "
          },
          "referenceSpecification": {
            "type": "Text",
            "value": "https://www.uco.es/rea/infor_rea/interpretacion.html"
          }
        }
      },
      "Fraxinus_Level": {
        "type": "Text",
        "value": "high",
        "metadata": {
          "description": {
            "type": "Text",
            "value": "Concentration level of gp per m3 - Allergenicity: A "
          },
          "referenceSpecification": {
            "type": "Text",
            "value": "https://www.uco.es/rea/infor_rea/interpretacion.html"
          }
        }
      },
      "Moraceae_Level": {
        "type": "Text",
        "value": "low",
        "metadata": {
          "description": {
            "type": "Text",
            "value": "Concentration level of gp per m3 - Allergenicity: A "
          },
          "referenceSpecification": {
            "type": "Text",
            "value": "https://www.uco.es/rea/infor_rea/interpretacion.html"
          }
        }
      },
      "Myrtaceae_Level": {
        "type": "Text",
        "value": "low",
        "metadata": {
          "description": {
            "type": "Text",
[...]

PR #226, opened 2018-03-01
Aero Allergen Observed data model draft

chicco785:
Provide a data model for Aero Allergen Observed as discussed in #221

### DIFF
--- a/Environment/AeroAllergenObserved/README.md
+++ b/Environment/AeroAllergenObserved/README.md
@@ -17,10 +17,10 @@ This entity models aero allergens observed at a given location and
   "id": "AeroAllergenObserved-CDMX-Pollen-Cuajimalpa",
   "type": "AeroAllergenObserved",
   "alnusLevel": "moderate",
-  "alnusConcentration": 40,
+  "alnus": 40,
   "alnusAllergenicity": "3",
   "casuarinaLevel": "low",
-  "casuarinaConcentration": 1,
+  "casuarina": 1,
   "casuarinaAllergenicity": "3",
   "allergenRisk": "moderate",
   "address": {
--- a/Environment/AeroAllergenObserved/README.md
+++ b/Environment/AeroAllergenObserved/README.md
@@ -1,0 +1,42 @@
+# Aero Allergen Observed
+
+## Description
+This entity models aero allergens observed at a given location and
+  related overall allergen risk.
+
+**Note**: JSON Schemas only capture the NGSI simplified representation, this
+  means that to test the JSON schema examples with
+  a [FIWARE NGSI version 2](http://fiware.github.io/specifications/ngsiv2/stable)
+  API implementation, you need to use the `keyValues`
+  mode (`options=keyValues`).
+
+## Examples of Use
+
+```json
+{
+  "id": "AeroAllergenObserved-CDMX-Pollen-Cuajimalpa",
+  "type": "AeroAllergenObserved",
+  "alnusLevel": "moderate",
+  "alnusConcentration": 40,
+  "alnusAllergenicity": "3",
+  "casuarinaLevel": "low",
+  "casuarinaConcentration": 1,
+  "casuarinaAllergenicity": "3",
+  "allergenRisk": "moderate",
+  "address": {
+    "addressCountry": "MX",
+    "addressLocality": "Ciudad de México",
+    "streetAddress": "Colegio Franco-Inglés"
+  },
+  "dateModified": "2018-02-16T17:24:39.00Z",
+  "dateObserved": "2018-02-11T00:00:00.00Z",
+  "location": {
+    "type": "Point",
+    "coordinates": [
+      -99.276977,
+      19.381877
+    ]
+  },
+  "source": "http://rema.atmosfera.unam.mx/rema/"
+}
+```
--- a/Environment/AeroAllergenObserved/README.md
+++ b/Environment/AeroAllergenObserved/README.md
@@ -16,12 +16,12 @@ This entity models aero allergens observed at a given location and
 {
   "id": "AeroAllergenObserved-CDMX-Pollen-Cuajimalpa",
   "type": "AeroAllergenObserved",
-  "alnusLevel": "moderate",
+  "alnus_Level": "moderate",
   "alnus": 40,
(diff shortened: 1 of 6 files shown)


jmcanterafonseca:
two comments:

In the Air Quality Data Model we are concatenating, terms using '_' as separator. Why don't we use the same patterns here, so that we would have something like:

````
{
   "alnus": 40,      /* This is the concentration */
   "alnus_Level": "moderate",
   "alnus_Allergenicity": 3
}
````

**Comment by jmcanterafonseca** [comment]
overall comment 2:

It would be good ro provide a list of the common allergen names that can be observed. or a link to some resource from the WHO or similar, as we did with the AirQualityObserved.

File: Environment/AeroAllergenObserved/README.md [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+## Examples of Use
+
+```json
+{
+  "id": "AeroAllergenObserved-CDMX-Pollen-Cuajimalpa",
+  "type": "AeroAllergenObserved",
+  "alnusLevel": "moderate",
+  "alnusConcentration": 40,
+  "alnusAllergenicity": "3",
```
jmcanterafonseca: why is this a number?  is it needed to be between quotes?

Environment/AeroAllergenObserved/doc/spec.md
```diff
@@ -0,0 +1,172 @@
...
+
++ Attribute type: [Number](https://schema.org/Number)
+
++ Attribute value: corresponds to the concentration of the allergen as a number.
+
++ Attribute metadata:
+    + `timestamp` : optional timestamp for the observed value in ISO8601 format.
+        It can be omitted if the observation time is the same as the one captured by the `dateObserved` attribute at entity level.
+        + Type: [DateTime](https://schema.org/DateTime)
+    + `unitCode` : The unit code (text) of measured concentration (usually grains per cubic meter).
```
jmcanterafonseca: is there a unit code for this measurement? can you put it as an example?

Environment/AeroAllergenObserved/doc/spec.md
```diff
@@ -0,0 +1,172 @@
...
+    + Mandatory
+    
++ `source` : A sequence of characters giving the source of the entity data.
+    + Attribute type: [Text](https://schema.org/Text) or [URL](https://schema.org/URL)
+    + Optional
+    
++ `allergenRisk` : Overall allergen risk corresponding to the aero allergens observed.
+    + Attribute type: [Text](https://schema.org/Text)
+    + Example values defined by the [European Aeroallergen Network](https://www.ean-net.org/en/):
+  (`none`, `low`, `moderate`, `high`, `very high`).
```
jmcanterafonseca: better `veryHigh`

Environment/AeroAllergenObserved/doc/spec.md
```diff
@@ -0,0 +1,172 @@
...
++ `refDevice` : A reference to the device(s) which captured this observation.
+    + Attribute type: Reference to an entity of type `Device`
+    + Optional
+
+### Representing aero allergens concentration
+
+In order to enable a proper management of the the different aero allergens concentration,
+*for each* aero allergens we have to use an attribute that *MUST* refers exactly
+to the conventional name of the allergen (usually the latin name of
+the associated plant) concatenated with the string `Concentration`,
```
jmcanterafonseca: as I said before we don't need to concatenate the word 'concentration', just use, as convention, the name of the allergen as the key

Environment/AeroAllergenObserved/doc/spec.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Representing qualitative levels of aero allergens
+
+In order to enable a proper management of the aero allergens qualitative levels,
+*for each* aero allergens we have to use an attribute that *MUST* refers exactly
+to the conventional name of the allergen (usually the latin name of
+the associated plant) concatenated with the string `Level`,
+e.g. `alnusLevel` to measure
+the qualitative level for a given concentration of 
+[alder](https://en.wikipedia.org/wiki/Alder) pollen.
```
jmcanterafonseca: alder or alnus?

jmcanterafonseca:
hi @chicco785

We plan to close v1.0 of Data Models at the end of May, so that the data models can be progressively migrated to NGSI-LD, in parallel with the GSMA IoT Big Data work.

Do you think you can polish this work by mid May?

Thanks!

chicco785:
hi jose,
it will be done this week

fede

chicco785:
```{
   "alnus": 40,      /* This is the concentration */
   "alnus_Level": "moderate",
   "alnus_Allergenicity": 3
}```

i don't think introducing the _ is really compliant with camelCase spec.

Environment/AeroAllergenObserved/doc/spec.md
chicco785: alder is the common english name of alnus (the latin name used in botanical classification)

Environment/AeroAllergenObserved/README.md
chicco785: As by spec: As this can be different between countries, regulations or implementations, the set of allowed values will depend on the reference specification used.

Environment/AeroAllergenObserved/doc/spec.md
chicco785: gr/m3, added in in the text

Environment/AeroAllergenObserved/doc/spec.md
chicco785: fixed

jmcanterafonseca:
@chicco785 although I'm sympathetic to your point regarding camel case, in this particular case I think we need to make an exception to allow client applications to easily parse the attributes. If we don't introduce the '_' as separator we there will be problems when parsing the string, or at least the string parsing will be less robust and/or obscure. So, I think the pragmatic solution to the problem is to introduce the '_' char as separator between the aeroalergen name and the derived property.

jmcanterafonseca:
we need links from mkdocs.yml to this new data model

chicco785:
@jmcanterafonseca 
6f24ef3 7ec71de should fix last comments by you

PR #227, opened 2018-03-12
Remove TEF as provider and substitute it by FIWARE

jmcanterafonseca:
@chicco785 can you r?


chicco785:
LGTM, but is FIWARE the provider? Isn't he supposed to be the owner of the original service?

jmcanterafonseca:
the data provider is who harmonizes the data. the original source is in the `source` attribute

Issue #234, opened 2018-03-28
Discrepancies between WeatherObserved spec and example harversters


aarranz:
The `WeatherObserved` spec (and schema) documents the use of the `barometricPressure` attribute while the example harvesters (e.g. [`spain_weather_boserved_harvester.py`](https://github.com/Fiware/dataModels/blob/master/Weather/WeatherObserved/harvest/spain_weather_observed_harvest.py)) uses an `atmosphericPressure` attribute. So, there are two options:

- Leaving untouched the spec, so we have to fix the harvesters to use a `barometricPressure` attribute  instead of the `atmosphericPressure` one.
- Or renaming the `barometricPressure` attribute in the spec and schema to `atmosphericPressure`.

I can provide a PR once decided what option fits better.

jmcanterafonseca:
@aarranz the authoritative source is https://www.gsma.com/iot/wp-content/uploads/2016/06/CLP.26-v4.0.pdf ,

please could you make a PR so that there is alignment?

jmcanterafonseca:
@aarranz ping

aarranz:
Sorry for not answering. Yes, I will provide a PR.

jmcanterafonseca:
closing. this was fixed

PR #241, opened 2018-05-12
Use Atmospheric Pressure instead of Barometric Pressure

aarranz:
This PR fixes issue #234, that is, it replaces the `barometricPressure` attribute on the `WeatherObserved` model with the `atmosphericPressure` attribute, that is the attribute documented by GSMA. This PR also replaces other occurrences in the `DeviceModel` spec and in the `AirQualityObserved` example harvesters.


jmcanterafonseca:
LGTM, thanks for checking all occurrences!

PR #245, opened 2018-05-21
Fix noise level observed. remove measurand

jmcanterafonseca:


Glagnar:
This looks like it introduces breaking changes.

jmcanterafonseca:
@Glagnar we are removing the measurand attribute because it was useless. that was done in the past in AirQualityObserved as well. The problem is that measurand does not allow to do queries and filter out the data properly, as it is a plain text field. 

However, existing data publishers can still use measurand and keep it in their datasets. The only thing is that they would need to add the content of measurand also as individual attributes as requested by the new data model.

jmcanterafonseca:
no news from @chicco785 . could someone devote some cycles to review or should I just land? @fgalan maybe?

Environment/NoiseLevelObserved/doc/spec.md
```diff
@@ -60,33 +60,21 @@ It allows to convey, roughly, information about the precision of the measurement
...
-    
-+ In order to enable a proper management of the *historical evolution* of the different acoustic parameters,
-*for each* element described by the `measurand` array list there *MAY* be an attribute which name *MUST* be exactly equal to the
-measurand name described on the `measurand` array. The structure of such an attribute will be as follows:
-    + Attribute name: Equal to the name of the measurand, for instance `LAeq`.
-    + Attribute type: [Number](https://schema.org/Number)
-    + Attribute value: Exactly equal (same unit of measurement) to the value provided in the `measurand` array.
-    
+The number of acoustic parameters measured can vary. *For each* acoustic measurand there *MUST* be an attribute which name *MUST* be exactly equal to the acoustic measurand name, as follows:
++ Attribute name: Equal to the name of the measurand, for instance `LAeq`, `LAeq,d`. It must correspond to a term defined at [http://www.acoustic-glossary.co.uk/definitions-l.htm](http://www.acoustic-glossary.co.uk/definitions-l.htm).
```
chicco785: my observation is that the naming of measurand is not consistent with air quality where you separate the "generic" measure from the specific measurand using _
also, introducing special characters such as "," may generate issues in the context broker and in other backends.

Environment/NoiseLevelObserved/doc/spec.md
```diff
@@ -40,15 +40,15 @@ This entity is primarily associated with the Smart City and environment vertical
 It can be represented by an specific time instant or by an ISO8601 interval. As a workaround for
 the lack of support of Orion Context Broker for datetime intervals, it can be used two separate attributes: `dateObservedFrom`, `dateObservedTo`. 
     + Attribute type: [DateTime](https://schema.org/DateTime) or an ISO8601 interval represented as [Text](https://schema.org/Text). 
-    + Mandatory
+    + Optional
         
 + `dateObservedFrom` : Observation period start date and time. See `dateObserved`. 
     + Attribute type: [DateTime](https://schema.org/DateTime). 
-    + Optional
+    + Mandatory
```
manucarrace: Please, be careful with stating attributes as Mandatory. It makes the model less usable

Environment/NoiseLevelObserved/doc/spec.md
```diff
@@ -40,15 +40,15 @@ This entity is primarily associated with the Smart City and environment vertical
 It can be represented by an specific time instant or by an ISO8601 interval. As a workaround for
 the lack of support of Orion Context Broker for datetime intervals, it can be used two separate attributes: `dateObservedFrom`, `dateObservedTo`. 
     + Attribute type: [DateTime](https://schema.org/DateTime) or an ISO8601 interval represented as [Text](https://schema.org/Text). 
-    + Mandatory
+    + Optional
         
 + `dateObservedFrom` : Observation period start date and time. See `dateObserved`.
```
manucarrace: Regarding to dateObservedFrom and dateObservedTo, I would say that they can be optional but not mandatory. Observed values may have a timestamp metadata stating the date the observation was adquired. It could involve that dateObservedFrom and dateObservedTo attributes to be not so relevant. My proposal is to state them as Optional.

jmcanterafonseca: Be careful. Noise level measurement always is performed during an integration period, so it makes sense to make these parameters mandatory.

Environment/NoiseLevelObserved/doc/spec.md
jmcanterafonseca: I disagree. We need to have something mandatory, at least a minimum meaningful set of attributes. Otherwise cities would claim they are compliant with but without publishing anything relevant.

Environment/NoiseLevelObserved/doc/spec.md
jmcanterafonseca: @chicco785 FWIW the "," is allowed in attribute names. However, in the ETSI specification we have disallowed it, so using '_' as separator instead of ',' could make sense, even though we will be breaking the international convention used.

jmcanterafonseca: @manucarrace `measurand` has been removed from the model as per this PR

jmcanterafonseca:
@chicco785 I've updated the model. Please let me know. 
 @manucarrace are u happy with my response comments?

Environment/NoiseLevelObserved/doc/spec.md
```diff
@@ -40,15 +40,15 @@ This entity is primarily associated with the Smart City and environment vertical
 It can be represented by an specific time instant or by an ISO8601 interval. As a workaround for
```
fgalan: According to the changes introduced in https://github.com/Fiware/dataModels/pull/246, `dateCreated` and `dateModified` above in this file (around L14) should change "Optional" to "Read-Only. Automatically generated."

fgalan: (Maybe the change is already in master and it is a matter of just updateing this PR branch with master... not sure)

Environment/NoiseLevelObserved/doc/spec.md
```diff
@@ -60,33 +59,24 @@ It allows to convey, roughly, information about the precision of the measurement
     + Attribute type: [Text](https://schema.org/Text)
     + Allowed values: one of (`"0"`, `"1"`, `"2"`)
     + Optional
+
```
chicco785: if it's always a number, why not using number?

Environment/NoiseLevelObserved/schema.json
```diff
@@ -0,0 +1,49 @@
...
+          "type": "string",
+          "format": "date-time"
+        },
+        "refDevice": {
+          "$ref": "https://fiware.github.io/dataModels/common-schema.json#/definitions/EntityIdentifierType"
+        },
+        "refPointOfInterest": {
+          "$ref": "https://fiware.github.io/dataModels/common-schema.json#/definitions/EntityIdentifierType"
+        },
+        "refWeatherObserved": {
```
chicco785: weather? not documented in the specs.

also, sonometerClass is missing.

Environment/NoiseLevelObserved/doc/spec.md
jmcanterafonseca: because it is not semantically a number, but a class, so I prefer to keep it text. In addition we can model better the allowed values by enumeration

Environment/NoiseLevelObserved/schema.json
jmcanterafonseca: yup, thanks for noticing

jmcanterafonseca:
we got approval from @chicco785 @manucarrace last chance to make your word. Landing deadline Tuesday 12th 12:00

Environment/NoiseLevelObserved/doc/spec.md
manucarrace: Understood. I think it's a trade off between backward compatibility and meaningful attributes.

manucarrace:
LGTM

PR #254, opened 2018-07-03
FIX location fields along all models

fgalan:
This PR improves location field in all existing models in the following way:

* It expand the usage of type to ge:point, geo:line, geo:polygon and geo:box, when applicable.
* For geo:json, it defines which expecific types are valid (Point, LineString, MultiLineString, etc.), when applicable.

CC: @jmcanterafonseca @manucarrace @mrutid

### DIFF
--- a/Alert/doc/spec.md
+++ b/Alert/doc/spec.md
@@ -50,7 +50,7 @@ found [here](../schema.json).
     + Optional
 
 + `location` : Location of alert represented by a GeoJSON geometry.
-    + Attribute type: `geo:json`.
+    + Attribute type: any of the location formats supported by NGSIv2 (`geo:point`, `geo:line`, `geo:box`, `geo:polygon` or `geo:json`).
     + Normative References: [rfc7946](https://tools.ietf.org/html/rfc7946)
     + Mandatory if `address` is not present
 
--- a/Device/Device/doc/spec.md
+++ b/Device/Device/doc/spec.md
@@ -58,7 +58,7 @@ and which are not currently covered by the standard attributes defined by this m
     + Optional
     
 + `location` : Location of this device represented by a GeoJSON geometry of type point. 
-    + Attribute type: `geo:json`.
+    + Attribute type: `geo:point` or `geo:json` (of type `Point`).
     + Normative References: [https://tools.ietf.org/html/rfc7946](https://tools.ietf.org/html/rfc7946)
     + Optional.
     
--- a/Environment/AeroAllergenObserved/doc/spec.md
+++ b/Environment/AeroAllergenObserved/doc/spec.md
@@ -30,7 +30,7 @@ A JSON Schema corresponding to this data model can be found [here](http://fiware
     + Read-Only. Automatically generated.
 
 + `location` : Location of the aero allergens observation represented by a GeoJSON geometry. 
-    + Attribute type: `geo:json`.
+    + Attribute type: any of the location formats supported by NGSIv2 (`geo:point`, `geo:line`, `geo:box`, `geo:polygon` or `geo:json`).
     + Normative References: [https://tools.ietf.org/html/rfc7946](https://tools.ietf.org/html/rfc7946)
     + Mandatory if `address` is not defined. 
     
--- a/Environment/AirQualityObserved/doc/spec.md
+++ b/Environment/AirQualityObserved/doc/spec.md
@@ -22,7 +22,7 @@ A JSON Schema corresponding to this data model can be found [here](http://fiware
     + Read-Only. Automatically generated.
 
 + `location` : Location of the air quality observation represented by a GeoJSON geometry. 
-    + Attribute type: `geo:json`.
+    + Attribute type: any of the location formats supported by NGSIv2 (`geo:point`, `geo:line`, `geo:box`, `geo:polygon` or `geo:json`).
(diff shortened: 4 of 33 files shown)


jmcanterafonseca:
-1 to geo:point, geo:xxx. It goes against the design principles of the data models. http://fiware-datamodels.readthedocs.io/en/latest/guidelines/index.html#modelling-location . We want only to stick geo:json for interoperability reasons. 

-1 to add geometries at discretion instead of particularising for each data model.

fgalan:
> -1 to geo:point, geo:xxx. It goes against the design principles of the data models. http://fiware-datamodels.readthedocs.io/en/latest/guidelines/index.html#modelling-location . We want only to stick geo:json for interoperability reasons.

I understand the need for homogenization, but using geo:json for simple cases (as Point) is overkill and, at the same time, limits the flexibility in NGSIv2. For instance, compare:

```
"location": {
  "type": "geo:json",
  "value": {
    "type": "Point",
    "coordinates": [ "-3.23", "40.01" ]
  }
}
```

with just

```
"location": {
  "type": "geo:point",
  "value": "40.01", "-3.23"
}
```

The second is much simple, isn't it? ;)

> -1 to add geometries at discretion instead of particularising for each data model.

Not sure if I understand what you mean here... could you elaborate/expand, please?

mrutid:
I mostly agree with the homogeneous solution. But I've one doubt... for example how could we restrict a value to be just a POINT when the type is "geo:json", it is possible to do so in the schema?... or we will be able just to state it as documentation associated to the model?

fgalan:
As [I predicted](https://github.com/Fiware/dataModels/pull/255#issuecomment-402651404) merging PR #255 has created a conflict on this PR. Pity... :(

jmcanterafonseca:
@fgalan the docs build was failing due to a readthedocs change which does not allow to have the mkdocs.yml in the same directory as the content and it had the maximum priority to be fixed.

On the other hand there is no agreement on this PR. Actually I'm not inclined towards opening the door to other location formats different than GeoJSON, so that the migration path to NGSI-LD and the final aim of FIWARE, which is interoperability, are achieved.

fgalan:
> On the other hand there is no agreement on this PR. 

To be precise there is no agreement *yet* on this PR ;)

> Actually I'm not inclined towards opening the door to other location formats different than GeoJSON, so that the migration path to NGSI-LD and the final aim of FIWARE, which is interoperability, are achieved.

Even in the case other location format wouldn't be include in the data models, this PR also include a better description of which types of GeoJSON are allowed in each case. For instance, if the model says "Location of this device represented by a GeoJSON geometry of type point" then the model has to state that the only valid GeJSON type is "Point".

chicco785:
Even in the case other location format wouldn't be include in the data
> models, this PR also include a better description of which types of GeoJSON
> are allowed in each case. For instance, if the model says "Location of this
> device represented by a GeoJSON geometry of type point" then the model has
> to state that the only valid GeJSON type is "Point".
>

i agree this is valuable

alvarosainzpardo:
We at work are trying to use the Fiware integrated models in our projects and we also think that using geo:json is simply too much metadata for just a pair of coordinates, especially when the data is sent by a device. In that situations, you have to send as little data as possible.

jmcanterafonseca:
> We at work are trying to use the Fiware integrated models in our projects and we also think that using geo:json is simply too much metadata for just a pair of coordinates, especially when the data is sent by a device. In that situations, you have to send as little data as possible.

@alvarosainzpardo good to hear from you!

Honestly I don't think GeoJSON adds any overhead, come on

"type": "Point",
"coordinates": [x,y]

is that any overhead? Let's use an standard! 
....

alvarosainzpardo:
Let's calculate the overhead:
- typical coordinates: xx.xxx,yy.yyy = 13 caracters total
- the same in geojson: {"type":"Point","coordinates":[xx.xxx,yy.yyy]} = 46 caracters total

46/13 is 3,5 times bigger, it seems to me like a big overhead. We are talking here about devices, that need to run in many cases from batteries, for years, sending several measures each day. Every byte count to save battery.

In many industrial applications, they use binary formats, compact (and propietary) just to send as few bytes as possible. I dont mean going that way, only to let people choose to use as few characters in the text protocol as possible, because that makes a lot of sense for devices in many situations.

fgalan:
Appart from @alvarosainzpardo comments (that I fully support) regarding

> is that any overhead? Let's use an standard!

Note that NGSIv2 (where geo:point is perfectly legal) is also a *de facto* standard, used by many many many context-based applications around the world ;)

jmcanterafonseca:
> Appart from @alvarosainzpardo comments (that I fully support) regarding
> 
> > is that any overhead? Let's use an standard!
> 
> Note that NGSIv2 (where geo:point is perfectly legal) is also a _de facto_ standard, used by many many many context-based applications around the world ;)

Yes, and that's why it uses GeoJSON ... and, another point, if you really need to use geo:point why don't you just do the conversion at IoT Agent level?

fgalan:
> Yes, and that's why it uses GeoJSON

That's correct. NGSIv2 allows both geo:json and geo:point (in addition to geo:line, geo:box and geo:polygon).

To make my point clear, I'm not meaning to use geo:point **instead of** geo:json in data models. What I mean is to allow geo:point **in addition to* geo:json. Let's be flexible and let the user to decide which one fits to his use case.

fgalan:
This PR seems to be obsolete. Closing.

PR #255, opened 2018-07-05
new root folder for data models

caa06d9c:
Because of readthedocs requirements, data models put in new root. Latest successfully build was one month ago.


fgalan:
I don't recommend to do a massive move of datamodels files while there are PRs on the fly touching these models...

jmcanterafonseca:
thank you. better name `docs` as `specs` 

also we need to change the URL associated to all the JSON schema files, maybe that can be automated

chicco785:
beyond that, this is a project about documentation, so i am not sure what is the benefit / purpose of introducing a root folder

caa06d9c:
Sorry for the wrong explanation, data models put under root folder because of readthedocs requirements. So if model stays in common root, read the docs will fail during the build.  So latest successful build (https://fiware-datamodels.readthedocs.io/en/latest/) was one month ago. Original comment will be fixed. @jmcanterafonseca can u you provide an example where this path should be fixed?

jmcanterafonseca:
https://github.com/Fiware/dataModels/blob/master/Alert/alert-schema.json#L3

chicco785:
not sure it is a good idea to change with url in the json schemas, because they can be referenced somewhere.

README.md
```diff
@@ -47,32 +47,34 @@ mode (`options=keyValues`).
 Contributions should come in the form of pull requests. 
 
 New data models should be added under a folder structured as follows:
-- `NewModel/`
-  - `doc/`
-    - `spec.md`: A data model description based on the [data model template](datamodel_template.md), e.g. [spec.md of WeatherObserved](Weather/WeatherObserved/doc/spec.md).
-  - `README.md`: A summary file (as an extract from the spec file), e.g. [README.md of WeatherObserved](Weather/WeatherObserved/README.md)
-  - `schema.json`: The JSON Schema definition, e.g. [schema.json of WeatherObserved](Weather/WeatherObserved/schema.json)
-  - `example.json`: One or more JSON example file, e.g. [example.json of WeatherObserved](Weather/WeatherObserved/example.json)
+- `docs/`
```
jmcanterafonseca: this should be `specs`

mkdocs.yml
```diff
@@ -2,82 +2,82 @@ site_name: Fiware-DataModels
...
+      - WeatherForecast      : Weather/WeatherForecast/doc/spec.md
+      - WeatherAlert         : Weather/WeatherAlert/doc/spec.md
+  -  Waste Management:
+      - Introduction         : WasteManagement/doc/introduction.md
+      - WasteContainerIsle   : WasteManagement/WasteContainerIsle/doc/spec.md
+      - WasteContainerModel  : WasteManagement/WasteContainerModel/doc/spec.md
+      - WasteContainer       : WasteManagement/WasteContainer/doc/spec.md
 
-google_analytics: ['UA-83719923-1', 'fiware-datamodels.readthedocs.io']
+google_analytics: ["UA-83719923-1", "fiware-datamodels.readthedocs.io"]
```
jmcanterafonseca: is this change really needed?

caa06d9c: as described in rthd examples. not critical, could be revert

jmcanterafonseca:
LGTM

PR #267, opened 2018-07-16
Add i18N guidelines

jmcanterafonseca:


specs/guidelines.md
```diff
@@ -84,7 +84,23 @@ the latter would be the creation date of the (digital) entity representing the w
...
 of the digital entity was updated, as typically there might be delay,
-specially on IoT networks which deliver data only at specific timeslots. 
+specially on IoT networks which deliver data only at specific timeslots.
+
+## Internationalization (i18N)
+
+There can be certain entity attributes which content is subject to be internationalized. For instance, the description of a Point of Interest.
+The i18N guidelines for the FIWARE Data Models are as follows:
+
++ By default the value of an attribute subject to be internationalized shall be expressed in *British English* (`en-GB`).
```
jason-fox: This conflicts with the data model guidelines which states that American English is default.

specs/guidelines.md
```diff
@@ -84,7 +84,23 @@ the latter would be the creation date of the (digital) entity representing the w
...
++ For each language variant available for an attribute subject to be internationalized there shall be an additional
+entity attribute which name shall be in the form:
+
+`<Attribute_Name>_<Language_Tag>`  where `Language_Tag` shall be a language tag as mandated by [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646.txt).
+W3C provides guidelines on [how to use language tags](https://www.w3.org/International/articles/language-tags/).  
+
+Example:
+
+an attribute named `description_es` can be used to convey the value of a `description` attribute in Spanish.
```
jason-fox: How can I tell which languages are supported for each attribute if all variations are top level?

e.g.

```json
{
   "description":  "Something Colorful in English",
   "description_en-gb":  "Something Colourful in British English",
   "description_de-de":  "Etwas Buntes auf Englisch ",
   "description_extended" : "The Color Blue"
}
```

However `description_extended` is **NOT** an exonym, but another field - How can I identify that?

Maybe we should do something like `name`/`alternativeName` as defined in schema.org:

```json
{
   "description":  "Something Colorful in English",
   "alternativeDescription" : [
       {"en-gb":  "Something Colourful in British English"},
       {"de-de":  "Etwas Buntes auf Englisch "}
    ],
   "description_extended" : "The Color Blue"
}
```

specs/guidelines.md
jmcanterafonseca: mmm you are right we have said that for naming attributes we should use American English ... however this refers to the content of attributes, should we say just *in English* and stay neutral ?

specs/guidelines.md
jmcanterafonseca: the JSON-LD @context allows you to fix the issue you are mentioning, look at

https://www.w3.org/TR/json-ld/#string-internationalization 

essentially, more than one JSON term could refer to the same property at semantic level and could be tagged with a language name at @context level as well. 

On the other hand the application shall try to detect suffixes which are actually valid language tags ... and not try to detect only attribute prefixes ...

jason-fox: OK, then could we clarify the usage by linking to the json-ld spec? Or even better, copy a relevant example:

```
{
  "@context": {
    ...
    "ex": "http://example.com/vocab/",
    "@language": "ja",
    "name": { "@id": "ex:name", "@language": null },
    "occupation": { "@id": "ex:occupation" },
    "occupation_en": { "@id": "ex:occupation", "@language": "en" },
    "occupation_cs": { "@id": "ex:occupation", "@language": "cs" }
  },
  "name": "Yagyū Muneyoshi",
  "occupation": "忍者",
  "occupation_en": "Ninja",
  "occupation_cs": "Nindža",
  ...
}
```

I assume you could use the Paris exonyms for this if you like.

specs/guidelines.md
jason-fox: Why not be consistent and re-state **American English** and leave no room for doubt?

specs/guidelines.md
```diff
@@ -89,18 +89,29 @@ specially on IoT networks which deliver data only at specific timeslots.
 ## Internationalization (i18N)
 
 There can be certain entity attributes which content is subject to be internationalized. For instance, the description of a Point of Interest.
-The i18N guidelines for the FIWARE Data Models are as follows:
+The internationalization (i18N) guidelines for the FIWARE Data Models are defined as follows:
 
-+ By default the value of an attribute subject to be internationalized shall be expressed in *British English* (`en-GB`).
-+ For each language variant available for an attribute subject to be internationalized there shall be an additional
++ By default, the value of an attribute subject to be internationalized shall be expressed in **American English** (`en-US`).
```
jason-fox: I'm not so sure about **shall** its a bit too prescriptive - should the default **always** be in English?  Is this necessarily the case? 

Take the city of *Livorno* in Italy. There is an obscure historical English exonym *Leghorn* 

If we insist on English as  a default then we would get:

```json
{
    "name": "Leghorn",
    "name_it": "Livorno"
}
```

Assuming there is no exonym in Spanish,  a Spanish reader would get an English geography lesson for the place name.  The default should be:

> the common international name in latin script, usually expressed in **American English**

Hence the following would be correct:

### English has a common exonym

```json
{
    "name": "Brussels",
    "name_fr": "Bruxelles",
    "name_nl": "Brussel",
    "name_de" : "Brüssel"
}
```
### English has a uncommon exonym
```json
{
    "name": "Livorno",
    "name_en": "Leghorn"
}
```
### Romanization of Chinese Script 
```json
{
    "name": "Beijing",
    "name_zh": "北京"
}
```

### No English Exonym - use common name
```json
{
    "name": "Tallinn",
    "name_de": "Reval"
}
```

jmcanterafonseca: good points one more time. I will try to clarify

PR #268, opened 2018-07-16
Add Contact Point to POI

jmcanterafonseca:


jmcanterafonseca:
that could be the SOS post telephone number, couldn't be?

now landing ...

PR #275, opened 2018-08-15
Amend Alert Model

jason-fox:
* Added `naturalDisaster` category
* Amended list of subcategory types - added missing and new entries
* Updated Documentation


jmcanterafonseca:
good work, thanks

PR #277, opened 2018-08-29
Add FIWARE favicon

jason-fox:

PR #278, opened 2018-09-06
Add favicon

jason-fox:


jason-fox:
Rebased change based of #277

Issue #238, opened 2018-04-25
[TrafficFlowObserved]


chicco785:
while the specs refer to `occupancy` the example mentions `capacity`.
are they the same? what's the correct one in case?

jmcanterafonseca:
I've made a PR with a tiny fix. but I realised that no JSON Schema is available nor example.json.

jmcanterafonseca:
I've assigned Fede as he intended to work on providing the JSON Schema

jmcanterafonseca:
@chicco785 do you think you would be able to provide a JSON Schema?

chicco785:
on the stack for next week (hopefully)

jmcanterafonseca:
already fixed, closing

PR #297, opened 2018-10-04
add schema and example for TrafficObservedFlow

chicco785:
* fix datetime example in the doc
* use 2 space indent in the example
* provide complete schema


chicco785:
fix #238

PR #303, opened 2018-10-11
Fix npm audits package-lock

jmcanterafonseca:


jmcanterafonseca:
well, let's land this and wait for @chicco785 fix on the package that is really causing the vulnerability.

PR #304, opened 2018-10-11
first modification for the specs document examples format

Joudsi:
@jmcanterafonseca

### DIFF
--- a/specs/Environment/AirQualityObserved/doc/spec.md
+++ b/specs/Environment/AirQualityObserved/doc/spec.md
@@ -244,37 +244,39 @@ Normalized NGSI response
 
 Sample uses simplified representation for data consumers `?options=keyValues`
 
-    {
-      "id": "Madrid-AmbientObserved-28079004-2016-03-15T11:00:00",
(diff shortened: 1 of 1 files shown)


specs/Environment/AirQualityObserved/doc/spec.md
```diff
@@ -237,8 +239,10 @@ mode (`options=keyValues`).
...
 }
 ```
-    
-## Examples of use 2 (?options=keyValues simplified representation for data consumers)
+
+## key-value pairs Example
+
+Sample uses simplified representation for data consumers `?options=keyValues`
 
     {
```
jason-fox: Indent and missing fence block

jmcanterafonseca:
@jason-fox the NGSI-LD flavour of the examples will be added eventually (I'm contributing a script to help with that task). So your suggestion on example entitling could make sense. 

Wrt fencing it comes from the old content. Probably it is a good idea to align both examples.

@Joudsi could you deal with @jason-fox comments? thanks!

Issue #310, opened 2018-10-23
howto page rendering has issues


jmcanterafonseca:
@jason-fox  @chicco785 

https://fiware-datamodels.readthedocs.io/en/latest/howto/index.html

PR #311, opened 2018-10-23
Remove whitespace around codeblock fences 

jason-fox:
Fix for #310

### DIFF
--- a/specs/howto.md
+++ b/specs/howto.md
@@ -119,7 +119,7 @@ To clarify the difference among the two representation here is a complete
 example from the `AirQualityObserved` Data Model:
 
 * Normalised format:
-  ```json
+```json
   {
       "id": "Madrid-AmbientObserved-28079004-2016-03-15T11:00:00",
       "type": "AirQualityObserved",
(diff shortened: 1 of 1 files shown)


specs/howto.md
```diff
@@ -316,7 +316,7 @@ Examples of GeoJSON geometries are available
   [UN/CEFACT](https://www.unece.org/fileadmin/DAM/cefact/recommendations/rec20/rec20_rev3_Annex3e.pdf). 
   E.g.:
 
-  ```json
+```json
```
chicco785: so readthedocs does not support indentation for code blocks to align them with bullets... a pity

on the other side, i think we should add a short guide on how to local generate the readthedocs so to check that they work fine

chicco785:
ps: also this list is not rendered correctly: https://github.com/Fiware/dataModels/pull/311/files#diff-9308877b8fc4a861aa6d718690d52f82R72

jason-fox:
We could either manually add two spaces on the sub list or push #312 and never have the problem again. I would be happy to abandon this if the other PR is accepted.

jason-fox:
Withdrawn - fixed by #312

PR #313, opened 2018-10-29
Fix spec doc

Joudsi:
done with specs example format for the Environment  folder in the specs


jmcanterafonseca:
@Joudsi please can you check ? it seems the branch has some conflicts ...

PR #316, opened 2018-10-31
Fix spec doc- Parking - offStreetParking

Joudsi:
@jmcanterafonseca


specs/Environment/AirQualityObserved/doc/spec.md
```diff
@@ -177,8 +177,7 @@ means that to test the JSON schema examples with a
 API implementation, you need to use the `keyValues` mode (`options=keyValues`).
 
 ## Examples
-
-### Normalized Example
+### Normalized  Example
```
jmcanterafonseca: this change is not needed, isn't it?

jmcanterafonseca:
@Joudsi  could you check the parking data model text? if you render the markdown it seems there are some bullet points incorrectly rendered. 

also please use the husky hook for doing so just do

cd validator
npm install 

and the hook will execute before each commit

thanks

jmcanterafonseca:
@Joudsi  it seems this branch is totally wrong. please make a PR from a branch that is rebased and not merged from the current master. thanks

Issue #318, opened 2018-10-31
Add text linter to the CI process


jmcanterafonseca:
/cc @chicco785

PR #320, opened 2018-11-02
Updating text linting rules and scope

jason-fox:
* Added filters to avoid false positives
* Updated spelling mistakes and dead URLs.


jason-fox:
I'm planning to add document linting/ link checking to CI as proposed here: 

* https://github.com/Fiware/dataModels/issues/318

There are occasions where we would get false positives from textlint - e.g. where we really want to use a cliche in the text or really want to keep in a URL which textlint has problems with. Adding the rule filter enables us to bypass these problems so that a GREEN build will occur when the document is responding as we see fit.

Without rule filtering, all CI Builds would be RED since there is no method to switch off false positives.

Adding code comments to disable issues is common enough in standard code linters for this very purpose - its even more important for text as by definition text rules are more fuzzy.

PR #321, opened 2018-11-02
Fix #318 - Add link checking to CI

### LINKED ISSUE #318: Add text linter to the CI process

jason-fox:
- update travis to run `npm test`

This needs to be applied after #320 to work successfully.

### DIFF
--- a/specs/Environment/AeroAllergenObserved/doc/spec.md
+++ b/specs/Environment/AeroAllergenObserved/doc/spec.md
@@ -148,7 +148,7 @@ To describe the allergenicity category of aero allergens,
 *for each* aero allergens we use an attribute that *MUST* refers exactly
 to the conventional name of the allergen (usually the latin name of
 the associated plant) concatenated with the string `_Allergenicity`,
-e.g. `alnus_Allergenicity` to describe the allergenicity level of 
+e.g. `alnus_Allergenicity` to describe the allergenicity level of
 [alnus](https://en.wikipedia.org/wiki/Alder) pollen.
 
 + Attribute name: Equal to the name of the allergen plus the suffix `_Allergenicity`, for instance `alnus_Allergenicity`.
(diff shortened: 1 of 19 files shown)


jason-fox:
The reason for splitting this out is that there are several ways that `textlint` could be enforced.

Either we could add this to **Travis** and check the text/links on pull requests or we could add a pre-process step on **Husky** to check the links when files are updated. At the moment I've just added the code for the former - the latter would only be a one line change in `package.json` though:

```
"*.md": [
			"textlint",
                         "prettier --tab-width 4 --print-width 80 --write --prose-wrap always",
			"git add"
		]

```
However, `textlint` run locally can be problematic - sometimes the link checking times out for no reason - I haven't had this issue when [testing my changes](https://travis-ci.org/jason-fox/dataModels/builds/449416579) on Travis though.

I wanted to keep this separate from #320 in case we want to back it out for some reason.

.travis.yml
```diff
@@ -32,9 +32,10 @@ matrix:
         - docker run -d --name orion1 --link mongodb:mongodb -p 1026:1026 fiware/orion -dbhost mongodb
       before_script:
         - npm run lint -C ./validator
-        - node ./validator/validate.js -i common-schema.json -i geometry-schema.json -i specs/Weather/weather-schema.json -i specs/Alert/alert-schema.json -p specs -c true
       script:
-        - true
+        - travis_wait 10 npm test -C ./validator
```
jmcanterafonseca: why we are waiting here? I would prefer not to have CI that depends on waiting for a certain amount of time

@caa06d9c any ideas?

jason-fox: This isn't really a wait within the CI, it is the standard mechanism to avoid a Travis
[Time out](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) when no output is received over a period of time.

Link checking can take time and we don't want CI to fail because it hasn't completed yet.
If the step is taking more than 10 minutes (with no output) then the CI build **will timeout** and error. 10 minutes should be enough for now.

PR #322, opened 2018-11-02
Fix some StreetLighting doc details

aarranz:
This PR fixes some problems in the StreetLightning specifications and provides initial JSON schemas for the `StreetlightGroup` and `StreetlightModel` data models.

### DIFF
--- a/specs/StreetLighting/StreetlightGroup/schema.json
+++ b/specs/StreetLighting/StreetlightGroup/schema.json
@@ -46,11 +46,17 @@
             "properties": {
               "from": {
                 "type": "string",
-                "format": "date-optionally-yearless"
+                "oneOf": [
+                  {"format": "date"},
+                  {"format": "ext-date-yearless"}
+                ]
               },
               "to": {
                 "type": "string",
-                "format": "date-optionally-yearless"
+                "oneOf": [
+                  {"format": "date"},
+                  {"format": "ext-date-yearless"}
+                ]
               },
               "hours": {
                 "type": "string"
--- a/specs/StreetLighting/StreetlightGroup/schema.json
+++ b/specs/StreetLighting/StreetlightGroup/schema.json
@@ -89,10 +89,6 @@
         "activeProgramId": {
           "type": "string"
         },
-        "dateModified": {
-          "type": "string",
-          "format": "date-time"
-        },
         "description": {
           "type": "string"
         },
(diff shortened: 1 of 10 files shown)


specs/StreetLighting/StreetlightGroup/schema.json
```diff
@@ -0,0 +1,121 @@
...
+          "format": "date-time"
+        },
+        "switchingOnHours": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "from": {
+                "type": "string",
+                "format": "date-optionally-yearless"
```
jmcanterafonseca: probably here you need to use `oneOf` so that both a date and a `ext-date-yearless` can be used .

specs/StreetLighting/StreetlightGroup/schema.json
```diff
@@ -0,0 +1,121 @@
...
+        },
+        "illuminanceLevel": {
+          "type": "number",
+          "minimum": 0,
+          "maximum": 1
+        },
+        "activeProgramId": {
+          "type": "string"
+        },
+        "dateModified": {
```
jmcanterafonseca: I think this is already defined in the common schemas included

validator/libs/schema.js
```diff
@@ -69,6 +69,7 @@ module.exports = {
     const file = path.join(fullPath, fileSchema);
     const schema = openFile(file, "schema");
     const ajv = new Ajv(conf.ajvOptions);
+    ajv.addFormat("date-optionally-yearless", /^(\d\d\d\d|-)-[0-1]\d-[0-3]\d$/);
```
jmcanterafonseca: please provide a more robust Reg Exp

specs/StreetLighting/StreetlightGroup/schema.json
aarranz: Yes, you're right. I'm removing it from the `Streetlight`, `StreetlightGroup` and `StreetligthControlCabinet` schemas.

validator/libs/schema.js
aarranz: Ok, do yo mean something like this (removing support for dates using `/` and `. ` for separator): https://stackoverflow.com/questions/15491894/regex-to-validate-date-format-dd-mm-yyyy/15504877#15504877?

jmcanterafonseca: the format should be `yyyy-mm-dd`  I would be fine if you are able to avoid 35 as day number :) and 13 as month number. Leap years etc. it is not needed to check ... I believe.

PR #332, opened 2018-12-04
More attribute fixes

Joudsi:
all files should be modified with the proper source and dataProvider attributes

@jmcanterafonseca


Joudsi:
I just deleted the modified WeatherObserved/README.md file from the pull request, I added the original one from the master. 

There were one conflict in the Vehicle file, I solved it.

it should be fine now.



@jmcanterafonseca

PR #338, opened 2018-12-18
Extra Categories for POI

jmcanterafonseca:
@emastrosimone could you r? please


jmcanterafonseca:
@emastrosimone could you r?

thanks!

PR #346, opened 2019-01-08
New Geometry schema

jmcanterafonseca:


jmcanterafonseca:
@juliozinga can you have a look as well, as you opened a PR initially

jmcanterafonseca:
@chicco785 please could your r? we need to land this asap

specs/UrbanMobility/GtfsShape/schema.json
```diff
@@ -55,4 +43,4 @@
     "type",
     "location"
   ]
-}
+}
```
chicco785: missing CR

geometry-schema.json
```diff
@@ -1,91 +1,224 @@
...
-            "properties": {
-                "type": { "enum": [ "Point" ] },
-                "coordinates": { "$ref": "#/definitions/position" }
-            }
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://geojson.org/schema/Geometry.json",
+  "title": "GeoJSON Geometry",
+  "oneOf": [
+    {
+      "$id": "#Point",
```
chicco785: internal $id is not used in the official schema, the rather provide a specific document per type to point directly to it.

geometry-schema.json
```diff
@@ -1,91 +1,224 @@
...
+          "type": "array",
+          "minItems": 4,
+          "items": {
+            "type": "number"
+          }
+        }
+      }
+    },
+    {
+      "$id": "#Polygon",
```
chicco785: internal $id is not used in the official schema, the rather provide a specific document per type to point directly to it.

geometry-schema.json
```diff
@@ -1,91 +1,224 @@
...
+          "type": "array",
+          "minItems": 4,
+          "items": {
+            "type": "number"
+          }
+        }
+      }
+    },
+    {
+      "$id": "#LineString",
```
chicco785: internal $id is not used in the official schema, the rather provide a specific document per type to point directly to it.

geometry-schema.json
```diff
@@ -1,91 +1,224 @@
...
+          }
         }
-    ],
-    "definitions": {
-        "position": {
-            "description": "A single position",
+      }
+    },
+    {
+      "$id": "#MultiPoint",
```
chicco785: internal $id is not used in the official schema, the rather provide a specific document per type to point directly to it.

geometry-schema.json
```diff
@@ -1,91 +1,224 @@
...
+          "type": "array",
+          "minItems": 4,
+          "items": {
+            "type": "number"
+          }
+        }
+      }
+    },
+    {
+      "$id": "#MultiLineString",
```
chicco785: internal $id is not used in the official schema, the rather provide a specific document per type to point directly to it.

geometry-schema.json
```diff
@@ -1,91 +1,224 @@
...
+          "type": "array",
+          "minItems": 4,
+          "items": {
+            "type": "number"
+          }
+        }
+      }
+    },
+    {
+      "$id": "#MultiPolygon",
```
chicco785: internal $id is not used in the official schema, the rather provide a specific document per type to point directly to it.

specs/UrbanMobility/GtfsShape/schema.json
```diff
@@ -24,28 +26,14 @@
...
-              "properties": {
-                "type": {
-                  "type": "string",
-                  "enum": ["LineString"]
-                },
-                "coordinates": { "$ref": "http://json-schema.org/geojson/geometry.json#/definitions/lineString" }
-              }
+          "oneOf": [
+            {
+              "$ref": "http://geojson.org/schema/Geometry.json#LineString"
```
chicco785: this cannot be referenced using actual url

specs/UrbanMobility/GtfsShape/schema.json
```diff
@@ -24,28 +26,14 @@
...
-                    "type": "string",
-                    "enum": ["MultiLineString"]
-                  },
-                  "coordinates": {
-                    "type": "array",
-                    "items": { "$ref": "http://json-schema.org/geojson/geometry.json#/definitions/lineString" }
-                  }
-                }
-           }]
+              "$ref": "http://geojson.org/schema/Geometry.json#MultiLineString"
```
chicco785: this cannot be referenced using actual url

geometry-schema.json
jmcanterafonseca: yes, that's why I've opened https://github.com/geojson/schema/issues/10

specs/UrbanMobility/GtfsShape/schema.json
jmcanterafonseca: It depends on how you map that URL to an actual schema. I don't like to have individual schemas per type of GeoJSON Geometry I think it does not make sense

chicco785: URL is by definition something that you resolve :) So while I understand your point, I disagree to the "hack". The solution is getting approved the change in GeoJSON and meanwhile adhere to what they offer.

geometry-schema.json
chicco785: i like having $id, what I don't like is not hacking the GeoJSON schema ad hoc.

jmcanterafonseca: but that' something needed, if the GeoJSON Repo people do not respond we cannot be blocked forever ...

specs/UrbanMobility/GtfsShape/schema.json
jmcanterafonseca: well, they are URIs not URLs ... https://json-schema.org/latest/json-schema-core.html#rfc.section.8.2 so not necessarily dereferenceable

chicco785: look, I am not stubborn on this, I just believe that not creating misalignments is the good way to go :)

jmcanterafonseca:
@chicco785 can we just land this with the commitment that if finally our change is not accepted by the GeoJSON Community, it will be aligned with whatever they produce?

thanks!

PR #348, opened 2019-01-08
Moving Air Quality Threshold to unsupported

jmcanterafonseca:


jmcanterafonseca:
I've added a simple readme to inform people

PR #349, opened 2019-01-09
added the normalized example to the building folder

Joudsi:
@jmcanterafonseca

### DIFF
--- a/specs/Building/Building/doc/spec.md
+++ b/specs/Building/Building/doc/spec.md
@@ -127,6 +127,7 @@ Normalized NGSI response
         "value": 7
     }, 
     "occupier": {
+        "type": "Relationship",
         "value": [
             "9830f692-7677-11e6-838b-4f9fb3dc5a4f"
         ]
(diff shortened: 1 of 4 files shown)


specs/Building/Building/example-normalized.json
```diff
@@ -0,0 +1,152 @@
+{
```
jmcanterafonseca: apply here also the comments above

specs/Building/Building/doc/spec.md
```diff
@@ -104,6 +104,172 @@ API implementation, you need to use the `keyValues` mode (`options=keyValues`).
...
+    "floorsBelowGround": {
+        "value": 0
+    }, 
+    "description": {
+        "value": "Office block"
+    }, 
+    "floorsAboveGround": {
+        "value": 7
+    }, 
+    "occupier": {
```
jmcanterafonseca: I think we need "type": "Relationship" as well

PR #366, opened 2019-01-16
Fix date in Garden

jmcanterafonseca:


specs/KeyPerformanceIndicator/schema.json
```diff
@@ -0,0 +1,194 @@
...
+          "type": "string",
+          "format": "date-time"
+        },
+        "area": {
+          "type": "string"
+        }
+      }
+    }
+  ],
+  "allOf": [
```
jason-fox: Duplicated `allOf` attribute

Issue #137, opened 2017-07-21
404 error when validating schemas


osfrance:
Hi, 
When I try to validate : 
  https://raw.githubusercontent.com/Fiware/dataModels/master/Weather/WeatherObserved/schema.json 
with data : 
  https://raw.githubusercontent.com/Fiware/dataModels/master/Weather/WeatherObserved/example.json 
I had an 404 error in : 
  https://fiware.github.io/dataModels/common-schema.json#/definitions/Location-Commons 
in Line 46 : 
  "location": { "$ref": "http://json-schema.org/geojson/geometry.json#" }, 
the url **http://json-schema.org/geojson/geometry.json** not exist !

Thanks for Help,
regards,

jmcanterafonseca:
you need to have the geometry.json file locally

ilarimikkonen:
@jmcanterafonseca I noticed the same. Could you please explain how this is supposed to work? How can we serve the file local, if it has an url assigned as reference?

fgalan:
@ilarimikkonen I have the same problem (see https://github.com/FIWARE/dataModels/pull/373#issuecomment-479969230)

jmcanterafonseca:
see 
https://github.com/FIWARE/dataModels/pull/373#issuecomment-480210961 

@ilarimikkonen that's easy it is served locally because the id is the same ...

albertoabellagarcia:
1) thanks for the comments
2) This repo is currently frozen and no longer evolution will be held there. Currently, we deal with it here  https://github.com/smart-data-models 
3) now it validates 
https://raw.githubusercontent.com/smart-data-models/dataModel.Weather/master/WeatherObserved/example.json
against the schema
https://raw.githubusercontent.com/smart-data-models/dataModel.Weather/master/WeatherObserved/schema.json
If there is anything else please reopen this issue
Thanks

PR #373, opened 2019-02-04
Add JSON Schema for NGSIv2

jmcanterafonseca:


fgalan:
I'll have a look along with week. Thanks!

Entity-ngsiv2-schema.json
```diff
@@ -0,0 +1,185 @@
...
+  "$schema": "http://json-schema.org/schema#",
+  "id": "https://fiware.github.io/dataModels/Entity-ngsiv2-schema.json",
+  "title": "NGSIv2 Entity",
+  "description": "NGSIv2 Entity",
+  "definitions": {
+    "EntityIdentifier": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 256,
+      "pattern": "^[\\w\\-\\.\\{\\}\\$\\+\\*\\[\\]`|~^@!,:\\\\]+$"
```
fgalan: Could you describe this pattern in "natural lenguage", pls? :)

Entity-ngsiv2-schema.json
```diff
@@ -0,0 +1,185 @@
...
+    },
+    "Name": {
```
fgalan: This "Name" is for attribute names and metadata names?

Take into account the following sections in [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/stable):

"Field syntax restrictions"
"Attribute names restrictions"
"Metadata names restrictions"

For instance, "id" and "type" literals are forbidden, there is a max lenght, * is forbidden, etc.

fgalan: In addition, pattern should be the same that for entities id (currently it isn't)

Entity-ngsiv2-schema.json
```diff
@@ -0,0 +1,185 @@
...
+      "type": [
+        "number",
+        "boolean",
+        "array",
+        "object",
+        "null",
+        "string"
+      ]
+    },
+    "TypeValue": {
```
fgalan: I think we should distinguish between *entity*, *attribute*, *metadata* types given they have different constrains. In particular, the special attributes for dates and geo-location only applies to *attributes*, as far as I remember.

In addition, the following sections in the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/stable/) should be taking into account:

"Field syntax restrictions"

Entity-ngsiv2-schema.json
```diff
@@ -0,0 +1,185 @@
...
+    },
+    "GeoTypeValue": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "const": "geo:json"
+        },
+        "value": {
+          "$ref": "http://json-schema.org/geojson/geometry.json#"
```
fgalan: geo:point, geo:box, geo:line and geo:poligon are also valid types in attributes so they should be added to the JSON schema.

Entity-ngsiv2-schema.json
```diff
@@ -0,0 +1,185 @@
...
+        "id": {
+          "$ref": "#/definitions/EntityIdentifier"
+        },
+        "type": {
+          "allOf": [
+            {
+              "$ref": "#/definitions/Name"
+            },
+            {
+              "default": "Thing"
```
fgalan: We can also take into account the default for attribute types, as described in "Partial Representations". Would it be possible?

Entity-ngsiv2-schema.json
```diff
@@ -0,0 +1,185 @@
...
+          "type": "object",
+          "properties": {
+            "metadata": {
+              "$ref": "#/definitions/Metadata"
+            }
+          }
+        }
+      ]
+    },
+    "EntityFragment": {
```
fgalan: What about the keyValues an valude entity representations described in the "Simplified Entity Representation" section of the spec? Does this schema cover them?

If not, also fine with it (the normalized format is the most complete one, so it makes sense we focus on it) but it should be taken into account and documented as limitation is some way.

Entity-ngsiv2-schema.json
```diff
@@ -0,0 +1,185 @@
...
+    },
+    "DateTimeTypeValue": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "const": "DateTime"
+        },
+        "value": {
+          "$ref": "#/definitions/DateTimeValue"
```
fgalan: This is "ours"? Or it refers to an existing and standard JASON schema for ISO8601 formatted dates?

Entity-ngsiv2-schema.json
jmcanterafonseca: it was developed time ago but it matches the NGSIv2 specification

Entity-ngsiv2-schema.json
jmcanterafonseca: see above

Entity-ngsiv2-schema.json
jmcanterafonseca: I would refrain from that as they are not used in the FIWARE Data Models

Entity-ngsiv2-schema.json
jmcanterafonseca: as you say this is only covering the normalized format

Entity-ngsiv2-schema.json
fgalan: As far as a I understand this is a schema for NGSIv2 specification, idependently of its usage in FIWARE Data Models. Thus, as long as geo:point, geo:box, geo:line and geo:polygon are part of NGSIv2 specification, they should be inluded in the schema.

Entity-ngsiv2-schema.json
fgalan: Thanks for the clarification. Clarification about it should be included in the PR.

Entity-ngsiv2-schema.json
fgalan: You mean a comment above this one? Or a line above this one?

(Sorry a probably stupid question :) but I'm not so familiar with JSON Schema and problably I'm missing important links to external JSON spec. If it is the second option and you can tell me around which code line, great!)

Entity-ngsiv2-schema.json
fgalan: From NGSIv2 spec:

> Allowed characters are the ones in the plain ASCII set, except the following ones: control characters, whitespace, `&`, `?`, `/` and `#`.

We should ensure that pattern matches exactly this. I'm not sure... it is a bit cryptic.

Maybe a ASCII range based (i.e. based in `\xNN` sintax) would be clearer?

Entity-ngsiv2-schema.json
jmcanterafonseca: see https://github.com/Fiware/dataModels/pull/373/files#diff-c4a04684aa22645d0665ae4a1afc7cf0R88
and
https://github.com/Fiware/dataModels/pull/373/files#diff-c4a04684aa22645d0665ae4a1afc7cf0R107

fgalan: Thanks for the links!

Ok, I think I understand how it works. With `oneOf` you combine the basic type with the special types in each case (DateTimeTypeValue in the case of metadata and DateTimeTypeValue  and GeoTypeValue in the case of attributes).

Thus NTC with regards this comment (although the incoroporation of additional geo: types is pending in [another comment](https://github.com/Fiware/dataModels/pull/373#discussion_r255896352)).

(Btw looking into this I have realized there is a small documentation flaw in NGSIv2 spec. See https://github.com/telefonicaid/fiware-orion/issues/3434. To be fixed)

Entity-ngsiv2-schema.json
fgalan: Maybe what can be done (assuming JSON Schema allows some form of inheritance, which I don't know) is to define a a base Identifier type with the restrictions in "Field syntax restrictions" then derive from it an AttributeName (adding the "Attribute names restrictions") and a MetadataName type (adding "Metadata names restrictions").

Entity-ngsiv2-schema.json
fgalan: Maybe specifying it in the root "description" field and in the filename of the schema would suffice.

Entity-ngsiv2-schema.json
jmcanterafonseca: It would complicate the schema unnecessarily 

NTC

Entity-ngsiv2-schema.json
jmcanterafonseca: I think it is perfectly clear, RegExp are cryptic by themselves and you have to always check documentation

jmcanterafonseca: NTC

Entity-ngsiv2-schema.json
jmcanterafonseca: fixed https://github.com/Fiware/dataModels/pull/373/commits/4599e4602ea40bb543bff6338fed72ee99806b60

Entity-ngsiv2-schema.json
jmcanterafonseca: NTC

Entity-ngsiv2-schema.json
jmcanterafonseca: well, we are in the context of the DataModels, this is the DataModels project ... 

NTC

Entity-ngsiv2-schema.json
jmcanterafonseca: fixed https://github.com/Fiware/dataModels/pull/373/commits/4599e4602ea40bb543bff6338fed72ee99806b60

Entity-ngsiv2-schema.json
jmcanterafonseca: not at all :)

We have created some definitions for DateTimeValue and DateTimeTypeValue, taking advantage of the JSON Schema constructs ;) Hope this helps

https://github.com/Fiware/dataModels/pull/373/files#diff-c4a04684aa22645d0665ae4a1afc7cf0R66

Entity-ngsiv2-schema.json
fgalan: Something like `[\x20-xFF]` (just an example) is probably much less cryptic than the current one in the PR.

I'd suggest to do the following: let's see how is the equivalent regex in "the `\xNN` way" looks like, compare with the existing one, and choose. Makes sense?

Entity-ngsiv2-schema.json
fgalan: ...but you are stating that this is an schema for NGSIv2. Thus, I'm the schema **is not a valid one for NGSIv2** is defined in the current way.

Entity-ngsiv2-schema.json
fgalan: I see 

```
 "format": "date-time"
```

Is that date-time a JSON Schema construct? Do you have a link to the place where it is documented?

Entity-ngsiv2-schema.json
fgalan: If you want to use an incomplete schema then the following:

```
  "title": "NGSIv2 Entity. Normalized Format",
  "description": "NGSIv2 Entity",
```

should change to something like this:

```
  "title": "NGSIv2 Entity. Normalized Format. No geo:point, geo:box, geo:line, geo:polygon",
  "description": "NGSIv2 Entity. Normalized Format. No geo:point, geo:box, geo:line, geo:polygon",
```

and the name of the file should also change, for instance Entity-ngsiv2-partial-schema.json.

Entity-ngsiv2-schema.json
jmcanterafonseca: https://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1

Entity-ngsiv2-schema.json
jmcanterafonseca: Entity-ngsiv2-datamodels-schema.json

Entity-ngsiv2-schema.json
jmcanterafonseca: I'm not gonna change this for the sake of changing, it works, it has been there for a long time, no need to change. 

NTC

Entity-ngsiv2-schema.json
fgalan: Thanks for the link!

NTC

Entity-ngsiv2-datamodels-schema.json
```diff
@@ -0,0 +1,238 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "id": "https://fiware.github.io/dataModels/Entity-ngsiv2-datamodels-schema.json",
+  "title": "NGSIv2 Entity. Normalized Format for FIWARE Data Models",
+  "description": "NGSIv2 Entity. Normalized Format for FIWARE Data Models",
```
fgalan: The text is ok for title, but description should be more descriptive. Proposal:

"JSON Schema for NGSIv2 Entity with two restrictions. First, it only covers normalized format (keyValues and Values are not covered). Second, it only covers geo:json for entity location attributes (geo:point, geo:line, geo:box and geo:polygon are not covered). These restrictions are aligned with the usage of this JSON Schema for FIWARE Data Models."

Entity-ngsiv2-schema.json
fgalan: Ok to the file name. However, description should be more cleaer. I have done a proposal in a separate comment.

NTC (with regards this comments thread)

Entity-ngsiv2-datamodels-schema.json
```diff
@@ -0,0 +1,238 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "id": "https://fiware.github.io/dataModels/Entity-ngsiv2-datamodels-schema.json",
+  "title": "NGSIv2 Entity. Normalized Format for FIWARE Data Models",
+  "description": "NGSIv2 Entity. Normalized Format aligned with the usage given by FIWARE Data Models. Restrictions: geo:point, geo:line, geo:polygon not covered. keyValues representation not covered",
```
fgalan: Pls add geo:box also to the list of restrictions.

Entity-ngsiv2-datamodels-schema.json
```diff
@@ -0,0 +1,238 @@
...
+  "definitions": {
+    "Identifier": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 256,
+      "pattern": "^[\\w\\-\\.\\{\\}\\$\\+\\*\\[\\]`|~^@!,:\\\\]+$"
+    },
+    "Name": {
```
fgalan: As far I understand this "Name" is used for entity types, attribute names, attribute types and metadata names, metadata types.

I'd suggest to have a type for each kind, given that the restrictions are different. For instance, the current schema as far as I know doesn't preclude to use "id" as attribute name (which is something that the NGSIv2 spec forbids). In addition, it forbids to use "geo:distance" as metadata name (something that would be weird but that, being strict, the NGSIv2 spec allows).

I mean, I propose to have "AttributeName", "AttributeType", "MetadataName", "MetadataType". With regards to entity type (used in "EntityFragment") I think that the "Identifier" type will be fine.

This proposal would make the schema a bit longer, but not too much and we'll gain in clearness and accuracy.

(I may be wrong. I'm not familiar with all the subtle details of JSON Schema syntax :). In that case, just tell me).

jmcanterafonseca: > As far I understand this "Name" is used for entity types, attribute names, attribute types and metadata names, metadata types.
> 
> I'd suggest to have a type for each kind, given that the restrictions are different. For instance, the current schema as far as I know doesn't preclude to use "id" as attribute name (which is something that the NGSIv2 spec forbids). 


In addition, it forbids to use "geo:distance" as metadata name (something that would be weird but that, being strict, the NGSIv2 spec allows).
> 
> I mean, I propose to have "AttributeName", "AttributeType", "MetadataName", "MetadataType". With regards to entity type (used in "EntityFragment") I think that the "Identifier" type will be fine.
> 
> This proposal would make the schema a bit longer, but not too much and we'll gain in clearness and accuracy.
> 
> (I may be wrong. I'm not familiar with all the subtle details of JSON Schema syntax :). In that case, just tell me).

JMCF: You won't be able to create an attribute which name is  `id` as obviously your JSON will be invalid. Providing a rule for that would be overkilling. 

I'm in favour of keeping the simplicity of the current schema and avoid being too detailed, specially for maintenance reasons and to allow other people to easily understand the format

Entity-ngsiv2-datamodels-schema.json
jmcanterafonseca: fixed

jmcanterafonseca:
@fgalan we need to land this, are you sufficiently happy with the current result? I do think it is in good shape for landing

thanks!

fgalan:
Last thing I remember is that I provide some comments to improve the modularity of the schema. I haven't review it in detail but given that github doesn't show any new commit since that, I understand these comments haven't been addressed.

Instead of keep discussing in comments I think it would be more productive to do the modifications myself and post them in this PR. However, I'm facing a problem trying to use the schema to validate. In particular, I'm getting 

```
jsonschema.exceptions.RefResolutionError: HTTP Error 404: Not Found
```

If I remove the part related with `http://json-schema.org/geojson/geometry.json` it works.

```
$ diff -u ngsiv2-draft-schema.json ngsiv2-draft-schema.orig.json
--- ngsiv2-draft-schema.json	2019-04-04 18:14:47.617443622 +0200
+++ ngsiv2-draft-schema.orig.json	2019-04-04 18:14:15.709442591 +0200
@@ -88,6 +88,22 @@
         "value"
       ]
     },
+    "GeoTypeValue": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "const": "geo:json"
+        },
+        "value": {
+          "$ref": "http://json-schema.org/geojson/geometry.json#"
+        }
+      },
+      "required": [
+        "type",
+        "value"
+      ]
+    },
     "Metadata": {
       "type": "object",
       "properties": {
@@ -154,6 +170,9 @@
               "$ref": "#/definitions/DateTimeTypeValue"
             },
             {
+              "$ref": "#/definitions/GeoTypeValue"
+            },
+            {
               "$ref": "#/definitions/TypeValue"
             }
           ]
```

It can be checked (e.g. in a browser) that http://json-schema.org/geojson/geometry.json resolves to 404. Moreover, I have found some people complaining about the same at https://stackoverflow.com/questions/48553563/how-to-acess-other-reference-to-http-json-schema-org-geojson-geometry-json

So, is that URL correct in the schema in this PR? Should it be changed for other?

fgalan:
In fact, the same problem seems to be reported in this (pretty old) issue: https://github.com/FIWARE/dataModels/issues/137

jmcanterafonseca:
I always do the validation of schemas locally using ajv, that makes the trick

```
ajv compile -s myschema.json -r geometry.json -r common.json , etc. 
```

fgalan:
I have found cases ([here](https://gist.github.com/fgalan/cee93f557c8c1428a963adde2ad6970e)) which are correct according NGSIv2 specification but not to the JSON schema in this PR. Thus, I'll modify the JSON Schema to solve this limitation.

fgalan:
[This](https://gist.github.com/fgalan/e115e7179b6bc3b2215de4813075ba9e) is the schema that solves the problem (included along with the diff with the current schema in the PR).

As far as I checked it correctly validates all [the cases previous commented](https://gist.github.com/fgalan/cee93f557c8c1428a963adde2ad6970e). Apart of it, it fix some additional problems in the current file (in particular, it wasn't taken into account the dateExpires build it attribute).

Please use the file in the gist to upgrade the one in this PR (I think I cannot do it myself, I don't commit persmission of this branch).

I'd suggest to run a wider regression. If you point me the set of files to validate I can do it.

jmcanterafonseca:
once #346 lands we will resume this with the contribution from @fgalan

fgalan:
> once #346 lands we will resume this with the contribution from @fgalan

Ok, perfect.

In the meanwhile

> I'd suggest to run a wider regression. If you point me the set of files to validate I can do it.

if you could point me to that set of files, I can try to run a regreesion.

jmcanterafonseca:
@fgalan I think this is ready for landing? Please, could you have a final check?

Thanks!

PR #379, opened 2019-02-07
Add generation date to the LD generator

jmcanterafonseca:


tools/ldcontext_generator.py
```diff
@@ -177,7 +178,8 @@ def write_context_file():
     print('writing LD @context...' + ' size: ' + str(len(aggregated_context)))
 
     ld_context = {
-        '@context': aggregated_context
+        '@context': aggregated_context,
+        'generatedAt': datetime.now().isoformat()
```
caa06d9c: datetime.now().replace(microsecond=0).isoformat()

PR #391, opened 2019-03-20
P2P car rental data model

ilarimikkonen:
This PR introduces a renting related data model, p2p car rental. It's developed in co-operation with Shareit Bloxcar Oy. The aim is to provide basic datamodel, which can be modified as needed.

### DIFF
--- a/specs/Renting/P2PCarRental/README.md
+++ b/specs/Renting/P2PCarRental/README.md
@@ -1,7 +1,8 @@
 # P2P Car Rental
 
-This entity contains a description of a peer to peer rental car.
-It aims to describe basic charateristics of the vehicle and it's owner. In addition to this it has some other details, like price and availability.
+This entity contains a description of a peer to peer rental car. It aims to
+describe basic charateristics of the vehicle and it's owner. In addition to this
+it has some other details, like price and availability.
 
 **Note**: JSON Schemas only capture the NGSI simplified representation, this
 means that to test the JSON schema examples with a
@@ -12,7 +13,6 @@ API implementation, you need to use the `keyValues` mode (`options=keyValues`).
 
 ```json
 {
-...TBD
+    ...TBD
 }
 ```
-
--- a/specs/Renting/P2PCarRental/README.md
+++ b/specs/Renting/P2PCarRental/README.md
@@ -1,0 +1,18 @@
+# P2P Car Rental
+
+This entity contains a description of a peer to peer rental car.
+It aims to describe basic charateristics of the vehicle and it's owner. In addition to this it has some other details, like price and availability.
+
+**Note**: JSON Schemas only capture the NGSI simplified representation, this
+means that to test the JSON schema examples with a
+[FIWARE NGSI version 2](http://fiware.github.io/specifications/ngsiv2/stable)
+API implementation, you need to use the `keyValues` mode (`options=keyValues`).
+
+## Examples of use
+
+```json
+{
+...TBD
+}
+```
+
--- a/specs/Renting/P2PCarRental/doc/spec.md
+++ b/specs/Renting/P2PCarRental/doc/spec.md
@@ -1,0 +1,13 @@
+# P2PCarRental
+
+## Description
+
+This entity contains a description of a peer to peer rental car.
+It aims to describe basic charateristics of the vehicle and it's owner. In addition to this it has some other details, like price and availability.
+
+
+## Data Model
+
+This data model has been created in co-operation with share it blox car.
+
+Please see the example.json and schema.json
--- a/specs/Renting/P2PCarRental/doc/spec.md
+++ b/specs/Renting/P2PCarRental/doc/spec.md
@@ -27,9 +27,9 @@ A JSON Schema corresponding to this data model can be found [here](../schema.jso
 
 -   `vehicle_information` : Follows [VehicleModel](https://github.com/FIWARE/dataModels/tree/master/specs/Transportation/Vehicle/VehicleModel) more details how to use the model are there. How ever, some of the semantics when using this datamodel are different. Here is a explanation how those are used in this model:
 
-	- `ID`: Unique id of the vehicle, can be autocreated.
+	- `Id`: Unique ID of the vehicle, can be autocreated.
 	- `name`: Descriptive name of the vehile, like "Alfa Romeo Giulia Quadrifoglio".
-	- `URL`: Url containing information about the vehicle.
+	- `Url`: URL containing information about the vehicle.
 	- `type`: Mandatory, needs to be "VehicleModel".
 	- `vehicleType`: Enum for example "car".
 	- `brandName`: Brandname, not necessarily same as manufacturer.
--- a/specs/Renting/P2PCarRental/doc/spec.md
+++ b/specs/Renting/P2PCarRental/doc/spec.md
@@ -6,42 +6,55 @@ This entity contains a description of a peer to peer rental car.
 It aims to describe basic characteristics of the vehicle and it's owner. In addition to this it has some other details, like price and availability.
 We hope that this model receives contributions and will be developed further by the community.
 
-This data model has been created in co-operation with share it blox car.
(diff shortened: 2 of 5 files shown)


specs/Renting/P2PCarRental/README.md
```diff
@@ -0,0 +1,4 @@
+# P2P Car Rental
+
```
jmcanterafonseca: is this about renting or just another transportation option?

specs/Renting/P2PCarRental/doc/spec.md
```diff
@@ -0,0 +1,75 @@
...
+document don't open up the usage.
+
+A JSON Schema corresponding to this data model can be found
+[here](../schema.json). All the following entries are mandatory.
+
+-   `id` : Unique identifier.
+
+-   `type` : Entity type. It must be equal to `P2PCarRental`.
+
+-   `vehicle_model` : The model of the vehicle omitting make, for example "159"
```
jmcanterafonseca: please follow the general data model guidelines ... and the documentation template

specs/Renting/P2PCarRental/doc/spec.md
```diff
@@ -0,0 +1,75 @@
+# P2PCarRental
+
+## Description
+
+This entity contains a description of a peer to peer rental car.
```
jmcanterafonseca: this is just a vehicle and we already have a Vehicle model ...

specs/Renting/P2PCarRental/README.md
ilarimikkonen: this is about renting

specs/Renting/P2PCarRental/doc/spec.md
ilarimikkonen: I checked the existing vehicle model and it did not seem to have all the stuff we need here. I'll check again to see how we can use that existing vehicle model. ANy suggestions?

jmcanterafonseca:
why don't you just use the VehicleModel data model (proposing an extension) to capture the static properties of your vehicle and just define an Entity type named P2PCarRental with the specific renting needs?

ilarimikkonen:
> why don't you just use the VehicleModel data model (proposing an extension) to capture the static properties of your vehicle and just define an Entity type named P2PCarRental with the specific renting needs?

Ok, sounds like a plan. How do you see the use of "id"? now the rental datamodel has an "id" and then the VehicleModel specifies "id" as mandatory. Should we use same ID for both, like "Rental-car-8f7ewf88ejjnf993jfU83HDJ" ?

ilarimikkonen:
+ how do I reference this schema: https://github.com/FIWARE/dataModels/blob/master/specs/Transportation/Vehicle/VehicleModel/schema.json from my json schema?

 I don't understand how this: https://github.com/FIWARE/dataModels/blob/master/specs/Transportation/Vehicle/Vehicle/schema.json and this: https://github.com/FIWARE/dataModels/blob/master/specs/Transportation/Vehicle/VehicleModel/schema.json

play together

jmcanterafonseca:
be careful, I'm talking about adding extra properties to VehicleModel not to create another VehicleModel schema. 

They don't play together they are a different kind of Entity. From Vehicle you point to a Vehicle Model.

ilarimikkonen:
> be careful, I'm talking about adding extra properties to VehicleModel not to create another VehicleModel schema.
> 
> They don't play together they are a different kind of Entity. From Vehicle you point to a Vehicle Model.

Ok, got it. I'll get started so that I just include the current VehicleModel properties and then propose new ones bit later once I get this one sorted.

I notice that you do point from Vehicle to VehicleModel, but I dont understand how (this is my 1st data model, please be patient), since the ref seems to be like this:

"refVehicleModel": {"$ref": "https://fiware.github.io/dataModels/common-schema.json#/definitions/EntityIdentifierType" }

Is this meaning that in this case it points to a particular existing instance on VehicleModel?

jmcanterafonseca:
build broken

https://travis-ci.org/FIWARE/dataModels/jobs/515171826

specs/Renting/P2PCarRental/doc/spec.md
```diff
@@ -0,0 +1,128 @@
...
+-   `type` : Entity type. It must be equal to `P2PCarRental`.
+
+-   `vehicle_average_rating`: Float, range from 1-5. Peers can give a rating for the renting experience.
+	-   Attribute type: Float
+	-   Optional
+
+-   `timezone` : For example: "Europe/Helsinki".
+	-   Normative reference: [IANA](https://www.iana.org/time-zones) spec.
+
+-   `vehicle_information` : Follows [VehicleModel](https://github.com/FIWARE/dataModels/tree/master/specs/Transportation/Vehicle/VehicleModel) more details how to use the model are there. How ever, some of the semantics when using this datamodel are different. Here is a explanation how those are used in this model:
```
jmcanterafonseca: be careful, I would reference a Vehicle Model but not inline it

jmcanterafonseca:
> "refVehicleModel": {"$ref": "https://fiware.github.io/dataModels/common-schema.json#/definitions/EntityIdentifierType" }
> 
> Is this meaning that in this case it points to a particular existing instance on VehicleModel?

Yes it is. Essentially you are going to have an Entity of type `FleetVehicle`, category `carRental`, that is going to have a `refVehicleModel` pointing to a `VehicleModel` instance.

So, the approach is, make a PR that extends the VehicleModel data model. Then, make another PR that creates the FleetVehicle data model. 

So, you are far from the solution, yet :)

jmcanterafonseca:
build is broken and feedback has not been taken into account, please check all comments again

ilarimikkonen:
> build broken
> 
> https://travis-ci.org/FIWARE/dataModels/jobs/515171826

fix pushes, will hopefully resolve

ilarimikkonen:
> > "refVehicleModel": {"$ref": "https://fiware.github.io/dataModels/common-schema.json#/definitions/EntityIdentifierType" }
> > Is this meaning that in this case it points to a particular existing instance on VehicleModel?
> 
> Yes it is. Essentially you are going to have an Entity of type `FleetVehicle`, category `carRental`, that is going to have a `refVehicleModel` pointing to a `VehicleModel` instance.
> 
> So, the approach is, make a PR that extends the VehicleModel data model. Then, make another PR that creates the FleetVehicle data model.
> 
> So, you are far from the solution, yet :)

I think we have some fundamental misunderstanding here. Question: Are all NGSI datamodels designed to work with Context Broker and only context broker?

ilarimikkonen:
What I'm trying to also judge here is that there is an existing API producing JSON very close to the one model I initially created. Now if I try to reuse existing NGSI datamodel pieces, (Vehicle, VehicleModel) I need to rethink the whole model, as the existing API relates to *renting* and the fact that the object for rent happens to be a vehicle is coincidence. That is the reason I created new "Renting" section.

Is the fundamental NGSI philosophy that datamodel contributors should reuse as much existing datamodels as possible? What I mean here is that from effort perspective if there is an existing API producing JSON it is very easy to drop it to a tool and the tool spits out a matching schema. If the datamodel needs to be constructed from pieces of existing datamodels, the effort increases.

jmcanterafonseca:
> Is the fundamental NGSI philosophy that datamodel contributors should reuse as much existing datamodels as possible? 

Yes, it is, we want coherency, we are welcoming a lot new contributions but they shall be aligned

> What I mean here is that from effort perspective if there is an existing API producing JSON it is very easy to drop it to a tool and the tool spits out a matching schema. If the datamodel needs to be constructed from pieces of existing datamodels, the effort increases.

Yes, but also your visibility and the reuse and impact of your work ...

ilarimikkonen:
I'm planning on closing this pull request. Requirements have changed and I don't feel that this makes sense, as it looks like we do not have an active data source for this. @jmcanterafonseca if you still see value in this, I can proceed with this, but otherwise I'd close it.

jmcanterafonseca:
You can close. Thanks

PR #392, opened 2019-03-20
Feature/travis

jason-fox:
This PR does the following:

1) Amends the travis script to run three separate jobs (Python 2.7, Python 3.6  and Docs)
2) Moves the tests from `before_script` to `script` so tests fail rather than error
3) Upgrades the Markdown linting to fail on format errors.
4) Re-run prettier on missing files.
5) Updated `.textlintrc` to ignore some more files to get the build to pass.

### Justification

```text
http://apidirectory.connectedliving.gsma.com
http://apidirectory.connectedliving.gsma.com/api/air-quality-spain
http://apidirectory.connectedliving.gsma.com/api/weather-spain
```

are currently dead - it looks like GSMA are doing some restructuring - maybe we'll need to alter or delete the relevant links, but I've ignored for now so other PRs won't fail incorrectly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: python
+sudo: required
+dist: xenial
 cache: pip
 
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ jobs:
          - node ./validator/validate.js -i common-schema.json -i geometry-schema.json -i specs/Weather/weather-schema.json -i specs/Alert/alert-schema.json -p specs -c true
 
     - stage: test
-      name: "Unit Tests 3.6"
-      python: 3.6
+      name: "Unit Tests 3.7"
+      python: 3.7
       node_js: 10
 
       before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,21 @@
 language: python
 cache: pip
-python:
-  - 2.7
-  - 3.6
-install:
-  - pip install flake8
-before_script:
-  # stop the build if there are Python syntax errors, PEP8 violations, undefined names
-  - flake8 . --count --select=E,F821,F822,F823 --max-line-length=127 --show-source --statistics
-  # exit-zero treats all errors as warnings.  GitHub editor is 127 chars wide
-  - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-script:
-  - true  # add other tests here
 
-matrix:
+jobs:
   include:
-    - language: node_js
-      python:
-      node_js:
-        - 8
+    - stage: test
+      name: "Unit Tests 2.7"
+      python: 2.7
+      node_js: 10
+
+      before_install:
+        - pip install flake8
+        # stop the build if there are Python syntax errors, PEP8 violations, undefined names
+        - flake8 . --count --select=E,F821,F822,F823 --max-line-length=127 --show-source --statistics
+        # exit-zero treats all errors as warnings.  GitHub editor is 127 chars wide
+        - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
       install:
-      ## Data model validator
+        ## Data model validator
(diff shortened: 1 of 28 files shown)


.travis.yml
```diff
@@ -32,12 +27,51 @@ matrix:
...
+        - sudo apt-get install -y docker-ce
+        - docker run --name mongodb -d mongo:3.4
+        - docker run -d --name orion1 --link mongodb:mongodb -p 1026:1026 fiware/orion -dbhost mongodb
+      before_script:
+        - npm run lint -C ./validator
+       
+      script:
+         - node ./validator/validate.js -i common-schema.json -i geometry-schema.json -i specs/Weather/weather-schema.json -i specs/Alert/alert-schema.json -p specs -c true
+
+    - stage: test
```
jmcanterafonseca: why do we need this?

.travis.yml
```diff
@@ -32,12 +27,51 @@ matrix:
...
+      install:
+        ## Data model validator
+        - npm install -C ./validator
+        - sudo apt-get remove -y docker docker-engine docker.io
+        - sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv EA312927
+        - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+        - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+        - sudo apt-get -qq update
+        - sudo apt-get install -y docker-ce
+        - docker run --name mongodb -d mongo:3.4
```
jmcanterafonseca: we can move to 3.6

.travis.yml
```diff
@@ -1,26 +1,21 @@
...
-matrix:
+
+jobs:
   include:
-    - language: node_js
-      python:
-      node_js:
-        - 8
+    - stage: test
+      name: "Unit Tests 2.7"
```
jmcanterafonseca: once @caa06d9c PR #363  on harvesters land we could get rid of this as everything will be in Python 3.7 (I hope)

for the time being is ok as this PR will land before than @caa06d9c one ... I think

.travis.yml
```diff
@@ -32,12 +27,51 @@ matrix:
         - docker run -d --name orion1 --link mongodb:mongodb -p 1026:1026 fiware/orion -dbhost mongodb
       before_script:
         - npm run lint -C ./validator
+       
+      script:
+         - node ./validator/validate.js -i common-schema.json -i geometry-schema.json -i specs/Weather/weather-schema.json -i specs/Alert/alert-schema.json -p specs -c true
+
+    - stage: test
+      name: "Unit Tests 3.6"
+      python: 3.6
```
jmcanterafonseca: could it be 3.7?

caa06d9c: I use 3.7.2, but not all harvesters  use it

.travis.yml
caa06d9c: ok

.travis.yml
jason-fox: Previously the Documentation tests were subsumed into the `before_script` step of the validator tests. - i.e. `npm run lint -C ./validator` - however when this fails the CI *errors* rather than *fails* and the rest of the script does not run.

It is much easier to track down whether the JSON or Docs or harvesters are at fault if they are covered by separate test suites in separate test steps.

.travis.yml
jason-fox: Fixed  484c9a2

.travis.yml
jason-fox: Fixed 3ad53be + 20f48dd - Travis offers 3.7.1 (using an official workaround)

.travis.yml
jason-fox: Furthermore, because the linter has now been set up to *fail* on errors, Doc fails are more likely to be found sooner. - e.g. 1edda9f - There was a missing bracket on one of the new models.

PR #431, opened 2019-04-29
Update TrafficFlowObserved data model

neeeel:


specs/Transportation/TrafficFlowObserved/schema.json
```diff
@@ -74,6 +74,9 @@
             "backward"
           ]
         },
+        "vehicleType": {
```
jmcanterafonseca: you need to convey what the possible vehicle types are. Please check the VehicleMode schema for a list of values. 

In addition you would need to modify the spec.md file to reflect the new attribute

thanks

neeeel: Hi, I have been looking into the vehicle types that are provided under the Vehicle Model, where it says the allowed values are defined by VehicleTypeEnum and VehicleTypeEnum2, DATEX 2 version 2.3

the allowed values ( agriculturalVehicle, bicycle, bus, minibus, car, caravan, tram, tanker, carWithCaravan, carWithTrailer, lorry, moped, tanker, motorcycle, motorcycleWithSideCar, motorscooter, trailer, van, caravan, constructionOrMaintenanceVehicle) do not necessarily easily fit into our vehicle classification ( LGV, OGV1, OGV2, car, etc). How are the allowed values maintained or changed, and how would I go about getting some values added?

neeeel: just to let you know, I have contacted synchronicity and they have answered the above question for me, I will look to updating the model and issuing a new pull request

specs/Transportation/TrafficFlowObserved/doc/spec.md
```diff
@@ -19,6 +19,15 @@ The data model is defined as shown below:
     -   Attribute type: Text or URL
     -   Optional
 
+-   `vehicleType` : Type of vehicle from the point of view of its structural characteristics.
+    -   See definition at [Vehicle](../../Vehicle/doc/spec.md).
+    -   Attribute metadata:
+        -   `subType` : A string value that allows more granular breakdown of the allowed vehicleType. Eg if the vehicleType is
+            set to `Lorry`, the subType may be `OGV1` or `OGV2`(specific types of lorry, specifying further characteristics not 
+            conveyed by the overall vehicleType)
+    -   Mandatory
```
jmcanterafonseca: this should be optional

specs/Transportation/TrafficFlowObserved/doc/spec.md
```diff
@@ -19,6 +19,15 @@ The data model is defined as shown below:
     -   Attribute type: Text or URL
     -   Optional
 
+-   `vehicleType` : Type of vehicle from the point of view of its structural characteristics.
+    -   See definition at [Vehicle](../../Vehicle/doc/spec.md).
+    -   Attribute metadata:
+        -   `subType` : A string value that allows more granular breakdown of the allowed vehicleType. Eg if the vehicleType is
```
jmcanterafonseca: do you think that metadata is a good solution for the subType?

Why don't we define a new attribute `vehicleSubType`

specs/Transportation/TrafficFlowObserved/schema.json
```diff
@@ -74,6 +74,38 @@
...
+            "trailer",
+            "van",
+            "caravan",
+            "constructionOrMaintenanceVehicle",
+            "trolley",
+            "binTrolley",
+            "sweepingMachine",
+            "cleaningTrolley"
+          ],
+          "metadata":{
```
jmcanterafonseca: be careful, schemas only define attribute values not metadata. that's why I'm suggesting that you define `vehicleSubType` as a new attribute

jmcanterafonseca:
@neeeel your PR needs to be in green state if it is not you need to investigate until it is, thanks!

specs/Transportation/TrafficFlowObserved/schema.json
neeeel: I was going on advice from synchronicity helpdesk with regards to using metadata to solve my problem. There is no fixed set of values for the vehicle sub type, so using metadata to specify a value seemed to make sense to me. I am open to advice from you about where we go, as I am fairly new to the fiware models and approach.

jmcanterafonseca: metadata is for data that refers to an Attribute, such as timestamp, units, etc. in this case it is not so clear ... so that's why I'm advising to model it as a new Attribute

neeeel: will it need a specified set of values, similar to how the vehicleType does?

specs/Transportation/TrafficFlowObserved/doc/spec.md
neeeel: @jmcanterafonseca , is there any way we can discuss this or resolve this directly, communicating via message board and then waiting for a reply is not working. I need to get this resolved as soon as I can.

Defining vehicleSubType as a new attribute may work, but would we need to define the acceptable values for this attribute, as happens with the vehicleType attribute?

neeeel: Hi @jmcanterafonseca , I have added in the vehicleSubType as a string attribute, and checked the commits against Travis CI, it fails one test, but the failure is a dead link in a different model, and so is nothing to do with my commits. 
thanks for your help

jmcanterafonseca:
@neel thanks. please could you rebase your PR so that we ensure the CI is green? thx!

```
git fetch upstream
git rebase upstream/master

git push origin master
```

specs/Transportation/TrafficFlowObserved/doc/spec.md
neeeel: @jmcanterafonseca sorry for my noobiness, not used to working with merges etc. Hopefully I have done it correctly now, I think I have added the changes to the correct branch by rebasing?

Issue #426, opened 2019-04-26
Add seeAlso to Agri entities so that they can be linked to other data model spaces


jmcanterafonseca:
seeAlso attribute Array of URIs

fdelavega:
I guess this is done

jmcanterafonseca:
fixed

PR #440, opened 2019-04-30
Add seeAlso attribute to AgriPest data model

fdelavega:
issue #426


jmcanterafonseca:
@fdelavega we can have in common-schema.json a new property common to any data model, so that seeAlso can be used in any of them. however, I see value in keeping the examples provided by this PR

PR #441, opened 2019-05-03
Proposal for `ThreePhaseMultiCircuitAcMeasurement`

brahma-dev:
A ThreePhaseMultiCircuitAcMeasurement entity represents a measurement from an electrical sub-metering system that monitors three-phase alternating current across multiple circuits. Each circuit is assigned a voltage phase: L1, L2 or L3; and has attributes for various electrical measurements such as `current`, `power`, `energy`, `frequency`, and `harmonics`.

### DIFF
--- a/specs/Energy/ThreePhaseMultiCircuitAcMeasurement/example-normalized.json
+++ b/specs/Energy/ThreePhaseMultiCircuitAcMeasurement/example-normalized.json
@@ -1,0 +1,96 @@
+{
+    "id": "ThreePhaseAcMultiCircuitMeasurement:LV3_Ventilation",
+    "type": "ThreePhaseAcMultiCircuitMeasurement",
+    "current": {
+        "value": [
+            33.55, 
+            34.77, 
+            31.32, 
+            32.4, 
+            34.08, 
+            13.33, 
+            10.86, 
+            11.11, 
+            23.31, 
+            11.31, 
+            14.16, 
+            40.19, 
+            29.6, 
+            13.36, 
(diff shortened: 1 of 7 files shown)


specs/Energy/ThreePhaseMultiCircuitAcMeasurement/example-normalized.json
```diff
@@ -0,0 +1,96 @@
+{
+    "id": "ThreePhaseAcMultiCircuitMeasurement:LV3_Ventilation",
+    "type": "ThreePhaseAcMultiCircuitMeasurement",
+    "current": {
+        "value": [
+            33.55,
```
jmcanterafonseca: instead of providing a value for each circuit why don't you just create a new entity per circuit?

brahma-dev: @jmcanterafonseca We gave it a try, but we find `circuit` to be very tightly coupled with the device as well as other circuits on the same device. And so we don't see a benefit in breaking it down into a new entity.

I've come up with an alternative schema that one might find easier to use : https://github.com/LincWorld/dataModels/commit/4209aff7f5ce0c4b4d6c54934ace7c7b64e598f3
Let me know if you prefer it. I'll send a pull request for that.

jmcanterafonseca: @brahma-dev your new proposal sounds good to me. Please re-do this PR and then I will review

PR #442, opened 2019-05-06
Add seeAlso attribute definition to common-schema.json

fdelavega:
Issue #426

### DIFF
--- a/common-schema.json
+++ b/common-schema.json
@@ -55,6 +55,15 @@
               }
             ]
           }
+        },
+        "seeAlso": {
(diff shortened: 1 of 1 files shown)


common-schema.json
```diff
@@ -55,6 +55,15 @@
               }
             ]
           }
+        },
+        "seeAlso": {
```
jmcanterafonseca: I think it could benefit from being either a string URI or an array of them (with minItems 1)

@fdelavega please could you update it? thanks

Issue #429, opened 2019-04-26
[Agri data models] relatedSource attribute 


jmcanterafonseca:
We need a new Attribute to identify related sources of entity data (other apps or systems)

```
relatedSource: {
  "type": "Property",
  "value": [{
     "application": <URI pointing to an AgriApp Entity>
    "entityId": <String with the related entity id in the app above>
  }]
}
```

jmcanterafonseca:
I'm afraid entityId does not refer to an NGSI Entity Id but to an entity id in an external app, so I think we need to rename this to `applicationEntityId` and let it be a plain string

jmcanterafonseca:
need to finish with the animal data model #474

jmcanterafonseca:
closing

PR #450, opened 2019-05-08
Add relatedSource attribute to AgriCrop data model

fdelavega:
#429


jmcanterafonseca:
I would suggest we add this to all Agri Data models.

You can create a agri-common-schema.json for that purpose and modify a bit the validator to include this schema as well ... may be you can give it a try? @fdelavega

PR #452, opened 2019-05-09
Add a common schema for AgriFood entities

fdelavega:
#429


specs/AgriFood/agrifood-schema.json
```diff
@@ -0,0 +1,31 @@
...
+                    "items": [
+                        {
+                            "type": "object",
+                            "values": {
+                                "application": {
+                                    "type": "string",
+                                    "format": "uri"
+                                },
+                                "entityId": {
+                                    "type": "string"
```
jmcanterafonseca: there is an already existing definition for entity ids ... please check common-schema.json

jmcanterafonseca:
we need to change this as described here

https://github.com/FIWARE/dataModels/issues/429

jmcanterafonseca:
LGTM, thanks

PR #453, opened 2019-05-14
Proposal for `ThreePhaseMultiCircuitAcMeasurement`

brahma-dev:
A ThreePhaseMultiCircuitAcMeasurement entity represents a measurement from an electrical sub-metering system that monitors three-phase alternating current across multiple circuits. Each circuit is assigned a voltage phase: L1, L2 or L3; and has attributes for various electrical measurements such as `current`, `power`, `energy`, `frequency`, and `harmonics`.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -31,6 +31,7 @@ pages:
       - DeviceModel            : Device/DeviceModel/doc/spec.md
   - Energy:
       - ThreePhaseAcMeasurement      : Energy/ThreePhaseAcMeasurement/doc/spec.md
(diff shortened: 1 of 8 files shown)


brahma-dev:
@jmcanterafonseca I have sent the MR @ https://gitlab.com/synchronicity-iot/synchronicity-data-models/merge_requests/23, but I think this PR is good for Fiware as well. Otherwise I can close this PR.

jmcanterafonseca:
I would prefer to close this, not to have the same thing in two different repos

Issue #425, opened 2019-04-26
[Agri] Enhance Device to include implements in Agri


jmcanterafonseca:
https://github.com/ADAPT/ADAPT/blob/c38badbeb9049a817353b3d810845adf3e6ff0b1/source/ADAPT/Equipment/DeviceElementTypeEnum.cs

fdelavega:
Not sure all the categories included in the linked enum can be considered a device. Fo example, machine used for tractors would map with Vehicle data model

jmcanterafonseca:
done.

PR #468, opened 2019-05-27
Add new category types to Device data models

fdelavega:
#425

### DIFF
--- a/specs/Device/DeviceModel/doc/spec.md
+++ b/specs/Device/DeviceModel/doc/spec.md
@@ -47,10 +47,14 @@ The data model is defined as shown below:
         multimedia content such as audio, images, animation, video.
         [https://w3id.org/saref#Multimedia](https://w3id.org/saref#Multimedia)
     -   `implement`: A device used or needed in a given activity; tool, instrument, utensil, etc.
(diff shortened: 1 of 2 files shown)


jmcanterafonseca:
you would need to also update the spec.md and to provide links to ADAPT specifications, including the meaning of each new type of device

fdelavega:
Not so sure that there is an ADAPT specification I can reference more that the C# file you included in the issue. Al least I haven't found documentation in this regard

jmcanterafonseca:
@fdelavega you need to check IoF2020 UC1.4 input and the links to ADAPT https://github.com/ADAPT/ADAPT/tree/develop/source/ADAPT

fdelavega:
There is no class for describing the DeviceElementType elements more than the C# enum, in some cases I can point to the C# config class, (Like ImplementConfiguration.cs) but I don't think this is the proper link

jmcanterafonseca:
@fdelavega can we find other devices from SAREF4Agri? and complete this work?

https://mariapoveda.github.io/saref-ext/OnToology/SAREF4AGRI/ontology/saref4agri.ttl/documentation/index-en.html

thanks!

PR #498, opened 2019-07-05
Fix streetlight nulls

jmcanterafonseca:


jmcanterafonseca:
@aarranz please can you review?

PR #499, opened 2019-07-05
Smart spot null

jmcanterafonseca:


jmcanterafonseca:
@FRDavid please can you review?

`null` values are no longer allowed as NGSI-LD does not allow them.

PR #509, opened 2019-07-16
small changes in harvesters

caa06d9c:
- uvloop added (performance)
- weather_type for Portugal weather forecast extended with additional values
- timeout implemented properly
- docker images fixed


jmcanterafonseca:
@caa06d9c please could you check CI?

we need to land this ASAP v1.0 of Data Models has to be created Monday, Tuesday, at the latest.

PR #512, opened 2019-08-06
fix #282 - adding number as available type

### LINKED ISSUE #282: [WeatherObserved] visibility is ambiguous
Property `visibility` is documented as
```
Allowed values: One of (veryPoor, poor, moderate, good, veryGood, excellent)
```
But this is very ambiguous, no sensor outputs this kind of scale. This is a problem for me because my data source gives me visibility as length in meters that can be seen. I didn't find easily any tables that could categorize visibility in the required scale.

caa06d9c:
The main fix is for "visibility, "windDirection", I suppose, should be changed to a number too, an integer is not in NGSI

### DIFF
--- a/specs/Weather/weather-schema.json
+++ b/specs/Weather/weather-schema.json
@@ -54,25 +54,18 @@
               "type": "string"
             },
             "visibility": {
-              "anyOf": [
-                {
-                  "type": "string",
-                  "enum": [
-                    "veryPoor",
-                    "poor",
-                    "moderate",
-                    "good",
-                    "veryGood",
-                    "excellent"
-                  ]
-                },
-                {
-                  "type": ["number", "null"]
-                }
+              "type": "string",
+              "enum": [
+                "veryPoor",
+                "poor",
+                "moderate",
+                "good",
+                "veryGood",
+                "excellent"
               ]
             },
             "windDirection": {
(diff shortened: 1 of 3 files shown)


caa06d9c:
ldcontext_generator_schema.json was renamed to exclude it from harvesting by ldcontext_generator.py

specs/Weather/weather-schema.json
```diff
@@ -54,18 +54,25 @@
...
             "visibility": {
-              "type": "string",
-              "enum": [
-                "veryPoor",
-                "poor",
-                "moderate",
-                "good",
-                "veryGood",
-                "excellent"
+              "anyOf": [
```
jmcanterafonseca: the change to this file should go https://github.com/smart-data-models

specs/Weather/weather-schema.json
```diff
@@ -54,18 +54,25 @@
...
+                    "veryPoor",
+                    "poor",
+                    "moderate",
+                    "good",
+                    "veryGood",
+                    "excellent"
+                  ]
+                },
+                {
+                  "type": ["number", "null"]
```
jmcanterafonseca: null should be removed from all the Schemas, it is not allowed in LD

PR #523, opened 2019-08-21
Fix #522 : Updating references to refer to core context.

### LINKED ISSUE #522: Amend contradictory LD definitions to ensure Core context always wins
Based on previous discussions we need to update some of the `@context` for the data models to ensure _Core context always wins_.

See https://github.com/FIWARE/context.Orion-LD/issues/180 and https://github.com/FIWARE/dataModels/issues/511

Any common attribute such as `type`, `name` or `description` (as defined in the core NGSI-LD spec) must defined as using **https://uri.etsi.org/ngsi-ld/...** as a normative type. 

To maintain the equivalency with `schema.org`, the text of the specification will be amended to state something like (*equivalent to XXX*) in the same manner that `schema.org/Person` = `foaf:Person` 

> ### Person
> Thing > Person
> A person (alive, dead, undead, or fictional).
> 
> Canonical URL: http://schema.org/Person
> Equivalent Class: foaf:Person

jason-fox:
1.  Prefer core context definitios
    *  https://uri.etsi.org/ngsi-ld/name is used over schema.org
    *  https://uri.etsi.org/ngsi-ld/description is used over schema.org

    In both cases a link to normative reference remains.

2. Schema.org links for Attribute Types have been added.
    *  Text and URL link to schema.org
    *  name and description are of type **Text**
    *  address is of type **address**
    * geo:json is a GeoProperty with appropriate normative reference.


specs/AgriFood/Animal/doc/spec.md
```diff
@@ -84,19 +84,19 @@ The proposed model for the animal entity has the following properties:
             -   Type: DateTime
     -   Optional
 -   phenologicalCondition: Phenological condition of the animal
-    -   Attribute type: Text
+    -   Attribute type: [Text](https://schema.org/Text)
```
jmcanterafonseca: we are missing here, `Property` (along the document)

specs/Environment/AirQualityObserved/doc/spec.md
```diff
@@ -139,7 +139,8 @@ measurand, ex. CO. The structure of such an attribute will be as follows:
         -   Type: [Text](https://schema.org/Text)
         -   Mandatory
     -   `description` : short description of the measurand
-        -   Type: [Text](https://schema.org/Text)
```
jmcanterafonseca: please double check if this description applies here

specs/AgriFood/Animal/doc/spec.md
jason-fox: Fixed b64335a

specs/Environment/AirQualityObserved/doc/spec.md
jason-fox: The description of the measurand is still a description and the definition of `schema.org/description` still applies:

> A description of the item.

Yes this is metadata (or *Property-of-a-Property* in NGSI-LD terms), but the *Property-of-a-Property* still has the name `description` which would expand to `https://uri.etsi.org/ngsi-ld/description` under NGSI-LD so I think it is still valid.

jmcanterafonseca: you are totally right!

jmcanterafonseca:
LGTM, thanks!

PR #533, opened 2019-09-04
Fix AgriProductType (was AgriProduct in many places)

jmcanterafonseca:


jmcanterafonseca:
yes, that changed has to be done by fixing issue #534

Issue #534, opened 2019-09-04
ensure Relationship "agriProduct" is renamed to "hasAgriProductType"


jmcanterafonseca:

fdelavega:
This one is very related to #528

fdelavega:
I think this one can be closed

jmcanterafonseca:
@fdelavega can you reference the PR on which this was fixed?

fdelavega:
#533 and #539

Issue #528, opened 2019-09-04
In the Agri Data Models align Relationship names to be always verbs


jmcanterafonseca:
This implies changes also in the GSMA Repos.

jmcanterafonseca:
@fdelavega what is the status of this?

fdelavega:
The FIWARE part is ready and the PR to the GSMA is approved. Althought, we are waiting for response regarding the potential issue with the context

jmcanterafonseca:
as the FIWARE part is fixed, closing this.

PR #539, opened 2019-09-12
Update relationships in Agri models to use verb attributes

fdelavega:
Issues #528 and #534


jmcanterafonseca:
please rebase your work to fix the build, as now the problem with the URL has been fixed

Issue #525, opened 2019-08-28
In the AgriDataModels check the missing Entity Types and add them


jmcanterafonseca:

jmcanterafonseca:
we can close this

PR #540, opened 2019-09-12
[WiP] Add new AgriFertiliser model

fdelavega:
Issue #525 

This model is referenced but not defined in GSMA ones, more work is required to identify relevant fields

### DIFF
--- a/specs/AgriFood/AgriFertiliser/example-normalized-ld.jsonld
+++ b/specs/AgriFood/AgriFertiliser/example-normalized-ld.jsonld
@@ -1,0 +1,31 @@
+{
+  "@context": [
+    "https://schema.lab.fiware.org/ld/context",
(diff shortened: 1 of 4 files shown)


AgrDIN-L:
Regarding AgriFertiliser (or more generic AgriProduct): A Product is modeled in ISO11783 and in ADM-ADAPT as a generic class, which gathers everything that could be considered input/output of a field Operation. That means for us, that we can considered the creation of a Property called ProductType (AgriProductType) and define some values for the property (e.g. Fertiliser or CropNutrition, CropProtection, CropVariety, etc.) and we relate the property to the AgriProduct generic entity.

jmcanterafonseca:
yes, sounds better to me that to model fertiliser as an AgriProductType

thanks!

PR #549, opened 2019-10-10
Add versioning guidelines

jmcanterafonseca:


specs/guidelines.md
```diff
@@ -188,6 +188,13 @@ In case of doubt check the existing data models!
 -   `relativeHumidity`
 -   `temperature`
 
+## Versioning
+
+Even though the FIWARE Data Models Project is aiming at keeping backwards compatibility, Data Providers
+may tag Entities to indicate the Schema version used when defining the content of an Entity,
```
jason-fox: > Even though the FIWARE Data Models Project is aiming at keeping backwards compatibility ...

FIWARE Data Models Project aim to maintain backwards compatibility, however some incompatibilities will inevitably occur over time.  Data providers may choose to tag Entities with an additional `schemaVersion` attribute so that Data Consumers can behave accordingly. This aligns with the [https://schema.org/schemaVersion](https://schema.org/schemaVersion) Property definition.

PR #556, opened 2019-10-17
Sort JSON-LD @context keys. Timezone in generated at

jmcanterafonseca:


jmcanterafonseca:
@jason-fox thanks for your review. there was a mess in the PR due to a previous squash of spurious commits from CI processes in test

PR #559, opened 2019-10-22
Fixing ci

stbnrivas:
testing if I can fix the CI doc errors.


stbnrivas:
cancel I will open new only with master base only

PR #575, opened 2020-05-12
Incorret Word in a sentence [ OnStreet ]

lrmendes:
In "category" -> "offstreet", this attribute is writen that "The parking spot belongs to an onstreet parking site". I think there is an error in the word "onstreet". 
The correct word would be "offstreet" in that sentence?
It only this, Thanks for the great work with that documentation.


albertoabellagarcia:
1.- thanks for the PR
2.- As long as these repos are frozen we have corrected the reported error here https://github.com/smart-data-models/dataModel.Parking/blob/master/ParkingSpot/

